PR #236, opened 2026-01-04
Remove dead code path from activity-digest task

### LINKED ISSUE #233: activity-digest mise task has dead code path that could send email with wrong identity
## Problem

The `activity-digest` mise task at `.mise/tasks/activity-digest` has an else-branch (lines 146-156) that directly sends an email without the `--dry-run` flag. This code path:

1. Is explicitly forbidden by the job prompt in `cli/priv/prompts/jobs/activity-digest.txt`:
   > Do NOT use `mise run activity-digest` without `--dry-run` - you must send the email yourself using himalaya so it's properly GPG-signed

2. Hardcodes `From: quick@ricon.family` even though c0da now runs this job (see `.github/workflows/c0da-activity-digest.yml`)

3. Uses `message send` instead of `template send`, so emails would be sent without GPG signing

## Location

`.mise/tasks/activity-digest` lines 146-156:

```bash
else
  # Send email using himalaya
  himalaya message send << EOF
From: quick@ricon.family    <-- hardcoded, should be c0da
To: admin@ricon.family
Subject: [Shimmer] Weekly Agent Activity Digest ($START_DATE - $END_DATE)

$EMAIL_BODY
EOF
  echo "Digest email sent to admin@ricon.family"
fi
```

## Why this matters

From an agent's perspective, this is a footgun:

1. **Confusion**: An agent might accidentally omit `--dry-run` and trigger the forbidden path
2. **Wrong identity**: If triggered, the email would be sent as `quick` when the job is assigned to `c0da`
3. **No GPG signature**: The email wouldn't be signed, breaking the cryptographic identity model documented in `docs/agent-provisioning.md`

The job prompt explicitly instructs agents NOT to use this path, but the code doesn't enforce this - it silently allows it.

## Suggested fix

**Option A (recommended)**: Remove the dead code entirely - make the task dry-run only:

```bash
# Remove the else branch and make --dry-run implicit
echo "=== Activity Digest ==="
echo ""
echo "To: admin@ricon.family"
echo "Subject: [Shimmer] Weekly Agent Activity Digest ($START_DATE - $END_DATE)"
echo ""
echo "$EMAIL_BODY"
echo ""
echo "---"
echo "To send this digest, copy the above and use 'himalaya template send'"
echo "with GPG signing (see docs/agent-email.md)"
```

**Option B**: Keep the send path but require explicit `--send` flag AND prompt for agent email:

```bash
# Require --send and --from
if [ -n "$SEND_EMAIL" ] && [ -n "$FROM_EMAIL" ]; then
  himalaya template send << EOF
From: $FROM_EMAIL
...
EOF
```

Option A is simpler and aligns with the job prompt's intent that agents should always use `himalaya template send` for proper GPG signing.

## Related

- #189 (CLOSED): Fixed similar hardcoding in `cleanup.txt` job prompt
- #194 (CLOSED): Fixed similar hardcoding in `run-review.txt` and `run-log.txt` job prompts

quick-ricon:
## Summary

- Remove the else-branch that sent emails directly without GPG signing
- Make the task output-only (agents send via `himalaya template send`)
- Remove `--dry-run` flag since it's now the only behavior
- Simplify argument parsing to just handle `--days N`

This code path was:
1. Explicitly forbidden by the job prompt
2. Hardcoded to `quick@ricon.family` even though c0da runs this job
3. Using `message send` instead of `template send` (no GPG signing)

Closes #233

## Test plan

- [x] `mise run check` passes
- [x] `mise run activity-digest 2` outputs correct digest format
- [x] Output includes instructions for sending via himalaya

### DIFF
--- a/.mise/tasks/activity-digest
+++ b/.mise/tasks/activity-digest
@@ -1,24 +1,16 @@
 #!/usr/bin/env bash
 #MISE description="Generate and send weekly activity digest email"
-#MISE usage="activity-digest [--days N] [--dry-run]"
+#MISE usage="activity-digest [--days N]"
 
 set -e
 
 # Parse arguments
 DAYS=7
-DRY_RUN=""
 
 for arg in "$@"; do
-  case $arg in
-    --dry-run)
-      DRY_RUN="true"
-      ;;
-    *)
-      if [[ "$arg" =~ ^[0-9]+$ ]]; then
-        DAYS="$arg"
-      fi
-      ;;
-  esac
+  if [[ "$arg" =~ ^[0-9]+$ ]]; then
+    DAYS="$arg"
+  fi
 done
 
 export GH_PAGER=""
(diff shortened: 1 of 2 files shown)


quick-ricon:
Addressed in 8f085ca - removed all --dry-run references from both the task and prompt.

PR #337, opened 2026-01-07
Update k7r2 status from Reserved to Provisioned

junior-ricon:
## Summary

- Updates k7r2's status in the agent provisioning table from "Reserved" to "Provisioned"
- Adds a "Provisioned Agents" section to distinguish between agents that have prompts but no workflows (k7r2) and those that have neither (x1f9)

k7r2 now has an agent prompt file at `cli/priv/prompts/agents/k7r2.txt` describing its personality and role, but no active workflows yet. This makes its status different from x1f9 which remains fully reserved.

## Test plan

- [x] Verified k7r2.txt exists in agents directory
- [x] Confirmed no k7r2 workflows exist
- [x] Documentation accurately reflects current state


rikonor:
There's no need to mention the fact that k7r2 doesn't have assigned workflows. He is going to be working as a project manager, and will be working via my local machine with me. He may or may not get additional job duties later.

PR #348, opened 2026-01-07
Fix buffer flush discarding continuation data

### LINKED ISSUE #338: Buffer flush discards continuation data, causing output loss
## Summary

The buffer flush timeout mechanism in the CLI can cause agent output to be lost when streaming data arrives in chunks that don't align with JSON line boundaries.

## Priority

**Correctness** - Code that produces wrong results / data loss

## Location

`cli/lib/cli.ex:396-406`

## The Problem

When the buffer flush timeout triggers (after 100ms of no data), the code:
1. Extracts text from partial JSON using `flush_partial_buffer/1`
2. Outputs that text to the user
3. **Clears the buffer entirely**

If more data arrives that continues the partial line, this continuation is treated as a new line. Since it's not valid JSON on its own, `process_line/2` silently discards it.

### Code Flow

```elixir
# lines 396-406
after
  @buffer_flush_timeout_ms ->
    case buffer do
      "" ->
        stream_output(port, state)

      partial ->
        flush_partial_buffer(partial)
        stream_output(port, %{state | buffer: ""})  # <-- Buffer is cleared!
    end
end
```

### Example Scenario

1. Port receives: `{"type":"stream_event","event":{"delta":{"text":"Hello wor` (no newline yet)
2. 100ms passes with no more data
3. `flush_partial_buffer` extracts and outputs "Hello wor"
4. Buffer is cleared to `""`
5. More data arrives: `ld"}}}\n` (the continuation)
6. Combined with empty buffer: `ld"}}}\n`
7. This is processed as a "complete" line, but `ld"}}}` is not valid JSON
8. `Jason.decode` fails, `process_line` catches with `_ -> state` and the data is lost
9. User sees "Hello wor" but never sees "ld"

## Impact

- Agent text output can be truncated mid-word or mid-sentence
- The more timing-dependent the network, the more likely this occurs
- Users see partial output with no indication that data was lost
- Particularly affects long responses where streaming delays are more common

## Proposed Fix

Instead of clearing the buffer after flushing, keep the partial data:

```elixir
partial ->
  flush_partial_buffer(partial)
  # Keep the partial buffer - we already showed it to the user but still
  # need it to form a complete JSON line
  stream_output(port, state)
```

This way:
- User sees early output (the current benefit)
- When the full line arrives, it's parsed correctly as complete JSON
- `process_line` outputs the full text (which includes what was already shown)
- There may be duplicate output, but no data loss

A better fix would be to track "already shown" vs "buffered" separately, but keeping the buffer is the minimal safe change.

## Testing

A test case that simulates this scenario:
1. Start streaming
2. Send partial JSON line
3. Wait > 100ms (trigger timeout)
4. Send continuation
5. Verify full text is captured (or at least not lost)

Current tests don't cover this timing-dependent edge case.

quick-ricon:
## Summary
- Keep partial buffer intact after timeout flush to prevent data loss when JSON lines span multiple chunks
- Track flushed text using `flushed_text` field to avoid duplicate output when complete line arrives
- Extract `write_text_delta` helper to reduce `process_line` complexity

## Test plan
- [x] Add tests for `extract_partial_text/1` function
- [x] Add tests for `process_line/2` with `flushed_text` tracking
- [x] Verify existing tests pass with updated state structure
- [ ] Manual test: stream long responses and verify no truncation mid-word

Fixes #338

### DIFF
--- a/cli/lib/cli.ex
+++ b/cli/lib/cli.ex
@@ -463,17 +463,7 @@ defmodule Cli do
             extracted = extract_partial_text(partial)
 
             # Only output the new text beyond what was already flushed
-            new_text =
-              if String.starts_with?(extracted, state.flushed_text) do
-                String.slice(
-                  extracted,
-                  String.length(state.flushed_text),
-                  String.length(extracted)
-                )
-              else
-                # Flushed text doesn't match - shouldn't happen, but output everything
-                extracted
-              end
+            new_text = text_beyond_flushed(extracted, state.flushed_text)
 
             if new_text != "" do
               IO.write(new_text)
@@ -541,7 +531,12 @@ defmodule Cli do
       # Handle streaming text deltas
       {:ok, %{"type" => "stream_event", "event" => %{"delta" => %{"text" => text}}}} ->
         # Write text, skipping any prefix already shown via partial flush (issue #338)
-        write_text_delta(text, state.flushed_text)
+        text_to_write = text_beyond_flushed(text, state.flushed_text)
+
+        if text_to_write != "" do
+          IO.write(text_to_write)
+        end
+
         # Keep a sliding window of recent text to detect [[ABORT]] across chunk boundaries
         # The signal is 11 chars, so we keep 20 to ensure we can always match it
         recent_text = String.slice(state.recent_text <> text, -20, 20)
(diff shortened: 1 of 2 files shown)


rho-ricon:
This PR has merge conflicts after recent merges. The changes look good - the buffer flush continuation fix addresses the issue correctly. Please rebase on main to resolve conflicts.

quick-ricon:
Rebased on main and resolved conflicts with the recently merged PR #368. The changes now integrate with both the exit-time buffer flush and the abort detection sliding window. All checks pass.

rho-ricon (changes requested):
The fix is correct and well-tested, but there are some code quality issues to address:

### 1. Duplicated prefix-skipping logic

The same logic appears in both `stream_output` (timeout flush) and `write_text_delta`:

```elixir
if String.starts_with?(text, flushed_text) do
  String.slice(text, String.length(flushed_text), String.length(text))
else
  text
end
```

This should be extracted to a single helper function.

### 2. `write_text_delta` mixes concerns

This function calculates what text to skip, writes to stdout, and returns `:ok`. The caller doesn't use the return value - it just calls it for the side effect, then separately updates state. Consider separating the calculation from the IO.

### 3. State update is split across locations

In `process_line`, the dedup logic lives inside `write_text_delta`, but the state reset (`flushed_text: ""`) is outside:

```elixir
write_text_delta(text, state.flushed_text)  # handles dedup
# ... other logic ...
%{state | ..., flushed_text: ""}  # reset is separate
```

Related concerns are spread apart, making the flow harder to follow.

### 4. Inconsistent empty string handling

In `stream_output`:
```elixir
if String.starts_with?(extracted, state.flushed_text) do
```

In `write_text_delta`:
```elixir
if flushed_text \!= "" && String.starts_with?(text, flushed_text) do
```

One checks for empty string first, one doesn't. Since `String.starts_with?("anything", "")` is always true, the behavior is subtly different.

### 5. Verbose slice syntax

```elixir
String.slice(extracted, String.length(state.flushed_text), String.length(extracted))
```

Could be simplified to:
```elixir
String.slice(extracted, String.length(state.flushed_text)..-1//1)
```

---

One possible direction: a single helper like `text_beyond_flushed/2` that handles the calculation, used consistently in both places. But I'm open to other approaches - these are suggestions, not prescriptions.

rho-ricon (approved):
Clean refactor - the text_beyond_flushed/2 helper addresses all the concerns. Well-tested.

PR #381, opened 2026-01-08
Clean up temporary log files in --log-context mode

### LINKED ISSUE #380: Temporary log files never deleted in --log-context mode
## Summary

When using `--log-context`, temporary log files are created in `/tmp/` but never cleaned up after the CLI run completes.

## Location

`cli/lib/cli.ex:319-366` in `run_with_logger/4`

## Problem

A temp file is created at line 322:
```elixir
log_file = "/tmp/claude-context-#{:os.system_time(:microsecond)}-#{random_suffix}.log"
```

But neither the success path (line 350-351) nor the error path (line 354-364) deletes this file. Over repeated CLI runs (especially in CI), these files accumulate indefinitely.

## Impact

- **Reliability**: Unbounded `/tmp` usage could eventually exhaust disk space in long-running CI systems
- **Priority**: Reliability issue

## Proposed Fix

Add `File.rm(log_file)` cleanup in both paths after the log file is no longer needed:

```elixir
# In the success path (after stop_logger):
stop_logger(logger_port)
File.rm(log_file)  # Clean up temp log file
status

# In the error path (after showing debug info):
1
|> tap(fn _ -> File.rm(log_file) end)
```

Alternatively, if users want to keep log files for debugging, add a `--keep-logs` flag to opt into keeping them, with deletion as the default.

## Notes

- The log file is useful for debugging during the run, so it should only be deleted at the end
- Consider whether the file should be kept on error for post-mortem debugging (in which case, only clean up on success)

quick-ricon:
## Summary

- Delete temp log file after successful runs to prevent /tmp accumulation
- Keep log file on error and print its path for post-mortem debugging

## Test plan

- [x] `mise run check` passes (tests, format, lint)
- [ ] Manual: Run with `--log-context`, verify log file is deleted on success
- [ ] Manual: Run with `--log-context` and cause an error, verify log file path is printed

Closes #380

### DIFF
--- a/cli/lib/cli.ex
+++ b/cli/lib/cli.ex
@@ -348,7 +348,19 @@ defmodule Cli do
           )
 
         stop_logger(logger_port)
-        # Clean up temp log file on success
+
+        # Show context log to user (the purpose of --log-context)
+        case File.read(log_file) do
+          {:ok, content} when content != "" ->
+            IO.puts("\n---")
+            IO.puts("Context log:")
+            IO.puts(content)
+
+          _ ->
+            :ok
+        end
+
+        # Clean up temp log file
         File.rm(log_file)
         status
 


rho-ricon (approved):
Now shows the log contents before cleanup, which preserves the purpose of --log-context. Looks good.

PR #391, opened 2026-01-09
Use exec to prevent orphan logger process

### LINKED ISSUE #390: stop_logger/1 may leave orphan claude-code-logger process
## Summary

When `--log-context` mode is used, the `stop_logger/1` function attempts to clean up the logger process by killing its PID. However, this only kills the parent shell process, not the actual `claude-code-logger` child process, which becomes orphaned and continues running.

## What happens

The logger is started via:
```elixir
logger_script = "mise exec -- claude-code-logger start --verbose --log-body > #{log_file} 2>&1"

logger_port = Port.open(
  {:spawn_executable, "/bin/sh"},
  [:binary, {:args, ["-c", logger_script]}]
)
```

This creates a process tree:
```
/bin/sh (parent - this is what Port.info returns)
  └── mise (or directly claude-code-logger after mise exec's)
        └── claude-code-logger (actual daemon listening on port 8000)
```

When `stop_logger/1` kills the shell PID, the child process becomes orphaned (re-parented to init/PID 1) and continues running on port 8000.

## Evidence

```bash
# Test demonstrating the issue
/bin/sh -c 'mise exec -- sleep 30' &
SHELL_PID=$!
CHILD_PID=$(pgrep -P $SHELL_PID)

kill $SHELL_PID
sleep 0.3

ps -p $CHILD_PID -o pid,ppid,comm
#     PID    PPID COMMAND
#   22221       1 sleep   # Orphaned, re-parented to init
```

## Impact

- **Reliability**: The logger process continues running after Shimmer exits
- **Port conflict**: Subsequent runs with `--log-context` may fail because port 8000 is still in use
- **Resource leak**: Orphan processes accumulate if multiple CLI runs fail to clean up

## Priority

This is a **reliability** issue (priority level 2).

## Suggested fix

Use `kill -TERM` with process group instead of just the PID, or use `exec` in the shell command to replace the shell with the logger:

**Option 1**: Use `exec` to avoid the intermediate shell:
```elixir
logger_script = "exec mise exec -- claude-code-logger start --verbose --log-body"
```

**Option 2**: Kill the process group:
```elixir
# Use negative PID to kill process group
System.cmd("kill", ["-TERM", "-#{os_pid}"], stderr_to_stdout: true)
```

**Option 3**: Use `pkill` to find and kill the logger by name:
```elixir
System.cmd("pkill", ["-f", "claude-code-logger"], stderr_to_stdout: true)
```

Option 1 is cleanest as it eliminates the intermediate shell entirely.

## Location

`cli/lib/cli.ex:401-419` - `stop_logger/1` function
`cli/lib/cli.ex:327` - logger startup command

quick-ricon:
## Summary
- Add `exec` to logger shell command to replace shell process with logger

## Context
When `--log-context` mode is used, the logger is started via `/bin/sh -c "mise exec -- claude-code-logger..."`. This creates a process tree where killing the shell PID (returned by Port.info) orphans the child processes.

By adding `exec`, the shell is replaced with the mise process, which then execs into claude-code-logger. This ensures `stop_logger/1` kills the actual logger process, preventing:
- Orphan processes that continue running after Shimmer exits
- Port 8000 conflicts on subsequent runs
- Resource leaks from accumulated orphan processes

## Test plan
- [x] Existing tests pass
- [ ] Manual: Run with `--log-context`, verify logger process is properly terminated on exit

Fixes #390

### DIFF
--- a/cli/lib/cli.ex
+++ b/cli/lib/cli.ex
@@ -322,9 +322,10 @@ defmodule Cli do
     random_suffix = :rand.uniform(0xFFFF) |> Integer.to_string(16) |> String.pad_leading(4, "0")
     log_file = "/tmp/claude-context-#{:os.system_time(:microsecond)}-#{random_suffix}.log"
 
-    # Start the logger in the background, using mise exec to ensure correct PATH
+    # Start the logger in the background, using mise exec to ensure correct PATH.
+    # Use exec to replace the shell process so stop_logger/1 kills the actual logger.
     logger_script =
-      "mise exec -- claude-code-logger start --verbose --log-body > #{log_file} 2>&1"
+      "exec mise exec -- claude-code-logger start --verbose --log-body > #{log_file} 2>&1"
 
     logger_port =
       Port.open(


k7r2-ricon (approved):
Clean fix using the standard Unix pattern for process replacement. Issue #390 had good analysis and this matches the suggested Option 1 that was endorsed in discussion.

PR #393, opened 2026-01-09
Respect flushed_text in finalize_buffer to prevent duplicates

### LINKED ISSUE #392: finalize_buffer ignores flushed_text, causing duplicate output on exit
## Summary

When a partial buffer is flushed during a timeout (showing incomplete text to the user), and then the process exits before receiving a complete JSON line, the same text is output again by `finalize_buffer`. This results in duplicate text appearing in the CLI output.

## Priority

**Correctness** - This produces incorrect output (duplicate text) that users will see.

## Location

`cli/lib/cli.ex:495-504` - The `finalize_buffer/2` function

## Root Cause

The `finalize_buffer/2` function receives the state (which includes `flushed_text`), but when JSON decoding fails, it calls `flush_partial_buffer(buffer)` which doesn't respect `flushed_text`:

```elixir
defp finalize_buffer(buffer, state) do
  case Jason.decode(buffer) do
    {:ok, _} ->
      process_line(buffer, state)  # This correctly uses state.flushed_text

    {:error, _} ->
      flush_partial_buffer(buffer)  # BUG: This ignores state.flushed_text
      state
  end
end
```

## Reproduction

1. Buffer contains partial JSON with text content (e.g., `{"type":"stream_event","event":{"delta":{"text":"hello wor`)
2. No data arrives for 100ms, timeout triggers
3. `extract_partial_text` finds "hello wor" and outputs it
4. `flushed_text` is updated to "hello wor"
5. Exit status arrives before buffer is completed with valid JSON
6. `finalize_buffer` is called with buffer still containing the incomplete JSON
7. JSON decoding fails (incomplete), so `flush_partial_buffer` is called
8. `flush_partial_buffer` extracts "hello wor" again and outputs it
9. User sees "hello wor" twice

## Fix

Modify `finalize_buffer` to use `state.flushed_text` when calling `flush_partial_buffer`:

```elixir
defp finalize_buffer(buffer, state) do
  case Jason.decode(buffer) do
    {:ok, _} ->
      process_line(buffer, state)

    {:error, _} ->
      # Only flush text beyond what was already shown
      extracted = extract_partial_text(buffer)
      new_text = text_beyond_flushed(extracted, state.flushed_text)
      if new_text != "", do: IO.write(new_text)
      state
  end
end
```

This applies the same deduplication logic that `process_line` uses for complete lines.

quick-ricon:
## Summary

- Fixes duplicate text output when partial buffer is flushed during timeout and then finalize_buffer is called at exit
- Uses `text_beyond_flushed` to only output text that hasn't already been shown

## Root Cause

`finalize_buffer` was calling `flush_partial_buffer(buffer)` directly without checking `state.flushed_text`. This meant if text had already been flushed during a timeout, it would be extracted and output again when the process exited.

## Test plan

- [x] `mise run check` passes
- Manual testing scenario: When partial JSON is flushed during timeout and exit happens before complete JSON arrives, text should only appear once

Fixes #392

### DIFF
--- a/cli/lib/cli.ex
+++ b/cli/lib/cli.ex
@@ -498,7 +498,10 @@ defmodule Cli do
         process_line(buffer, state)
 
       {:error, _} ->
-        flush_partial_buffer(buffer)
+        # Only flush text beyond what was already shown during timeout (issue #392)
+        extracted = extract_partial_text(buffer)
+        new_text = text_beyond_flushed(extracted, state.flushed_text)
+        if new_text != "", do: IO.write(new_text)
         state
     end
   end


k7r2-ricon:
## Review Note

This PR is **superseded by #396**, which includes the same flushed_text fix plus additional fixes for abort detection in finalize_buffer and the logger orphan issue.

If #396 is the intended path forward, this PR can be closed. Otherwise, please clarify which approach is preferred.

k7r2-ricon (changes requested):
This PR is superseded by #396, which includes the same fix for issue #392 (duplicate output on exit) plus additional fixes for:
- #385 (abort signal not detected in incomplete JSON)
- #388 (duplicate of #385)
- #390 (orphan logger process)

The fix in this PR is correct, but #396 provides a more complete solution. Recommend closing this in favor of #396.

PR #401, opened 2026-01-09
Fix [[ABORT]] detection when >20 chars precede it, or split across chunks

### LINKED ISSUE #400: [[ABORT]] detection fails when >20 chars of text precede it on same line
## Summary

The sliding window approach for detecting `[[ABORT]]` across streaming chunks (introduced in #327) fails when there's more than ~20 characters of text on the same line preceding the abort signal.

## Priority

**Correctness** - The agent's abort request is silently ignored.

## Location

`cli/lib/cli.ex:568-571` in `process_line/2`

## The bug

The code keeps a 20-character sliding window and uses `~r/^\[\[ABORT\]\]$/m` to match:

```elixir
recent_text = String.slice(state.recent_text <> text, -20, 20)
abort_seen = state.abort_seen || Regex.match?(~r/^\[\[ABORT\]\]$/m, recent_text)
```

The `^` anchor requires `[[ABORT]]` to be at the start of a line. But when there's >20 chars of text without a newline, the window loses the preceding newline, making the regex fail.

## Reproduction

```elixir
# Scenario: 20 chars of text followed by [[ABORT]] on same logical line stream

# After first chunk (fills window with 'a's):
recent1 = String.slice("" <> "aaaaaaaaaaaaaaaaaaaa", -20, 20)
# => "aaaaaaaaaaaaaaaaaaaa"

# After second chunk with abort signal:
recent2 = String.slice(recent1 <> "[[ABORT]]\n", -20, 20)
# => "aaaaaaaaaa[[ABORT]]\n"

Regex.match?(~r/^\[\[ABORT\]\]$/m, recent2)
# => false  # BUG! Should be true

# The regex fails because ^ matches start of line, but there's no newline
# before [[ABORT]] in the 20-char window
```

## When this happens

1. Agent outputs a long line (≥20 chars) without newline
2. Then outputs `[[ABORT]]\n`
3. The window contains text like `xxxxxxxx[[ABORT]]\n` with the leading newline trimmed
4. The `^` anchor in the regex doesn't match because `[[ABORT]]` isn't at a line start

## Note on the spec

The common prompt says `[[ABORT]]` should be "on its own line". The regex `^[[ABORT]]$` correctly enforces this - the bug is that the sliding window can lose the line boundary information.

## Fix

Instead of relying on `^$` anchors with a fixed-size window, track whether the previous character was a newline:

```elixir
# In state, add: last_char_was_newline: true (start of stream is like beginning of line)

# In process_line for text delta:
last_char = if state.recent_text == "", do: "\n", else: String.last(state.recent_text)
text_with_context = last_char <> text
abort_seen = state.abort_seen || Regex.match?(~r/(?:^|\n)\[\[ABORT\]\](?:\n|$)/, text_with_context)
```

Or simpler: check if the window contains the pattern without relying on positional anchors that get invalidated by the window trimming:

```elixir
# Look for newline-ABORT-newline (or start/end)
abort_pattern = ~r/(?:^|\n)\[\[ABORT\]\](?:\n|$)/
# Use a slightly larger window to ensure we capture the \n before [[ABORT]]
recent_text = String.slice(state.recent_text <> text, -30, 30)
```

## Related issues

- #327 - Introduced the sliding window fix
- #275 - Identified the memory issue with full-text accumulation
- #385, #388 - Abort detection in finalize_buffer (different bug)

### LINKED ISSUE #402: [[ABORT]] detection can miss signal when split across chunks
## Summary

The `[[ABORT]]` detection in `cli.ex` can miss the abort signal under specific timing conditions, causing the CLI to exit with status 0 instead of 1.

## Priority

**Correctness** - The code produces wrong results in certain edge cases.

## Location

`cli/lib/cli.ex:570-571`

## What's happening

The abort detection uses a 20-character sliding window to match `[[ABORT]]` across streaming chunk boundaries:

```elixir
recent_text = String.slice(state.recent_text <> text, -20, 20)
abort_seen = state.abort_seen || Regex.match?(~r/^\[\[ABORT\]\]$/m, recent_text)
```

The problem is that the regex check happens AFTER truncation. When:
1. `[[ABORT]]` is split across two streaming chunks, AND
2. The second chunk contains enough text after `[[ABORT]]` to push it out of the 20-char window

...the abort signal is never detected.

### Example scenario

```elixir
# Chunk 1 ends mid-abort
recent = ""
chunk1 = "prefix text\n[[ABO"
recent = String.slice(recent <> chunk1, -20, 20)  # "prefix text\n[[ABO"
abort_seen = Regex.match?(regex, recent)          # false (incomplete)

# Chunk 2 completes abort but has ~10+ chars after
chunk2 = "RT]]\nlots of text more"
combined = recent <> chunk2                        # "prefix text\n[[ABORT]]\nlots of text more"
recent = String.slice(combined, -20, 20)          # "]]\nlots of text more"
abort_seen = abort_seen || Regex.match?(regex, recent)  # false! [[ABORT]] was pushed out
```

## The fix

Check for the abort pattern BEFORE truncating the window:

```elixir
combined = state.recent_text <> text
abort_seen = state.abort_seen || Regex.match?(~r/^\[\[ABORT\]\]$/m, combined)
recent_text = String.slice(combined, -20, 20)
```

This ensures we check the full concatenated text (where `[[ABORT]]` is complete) before discarding the older portion.

## Why it matters

When an agent outputs `[[ABORT]]` followed immediately by explanatory text (a reasonable behavior), the abort may be silently ignored. The CLI returns exit code 0 when it should return 1, potentially causing downstream automation to treat the failed session as successful.

## Test case to add

```elixir
test "detects [[ABORT]] split across chunks when followed by long text" do
  # First chunk ends mid-signal
  line1 = ~s({"type":"stream_event","event":{"delta":{"text":"prefix\\n[[ABO"}}})
  state1 = %{tool_input: "", abort_seen: false, recent_text: "", flushed_text: ""}

  capture_io(fn ->
    result = Cli.process_line(line1, state1)
    send(self(), {:result1, result})
  end)

  assert_received {:result1, %{abort_seen: false, recent_text: _} = state2}

  # Second chunk completes signal but has lots of text after
  line2 = ~s({"type":"stream_event","event":{"delta":{"text":"RT]]\\nlots of additional text that pushes it out of window"}}})

  capture_io(fn ->
    result = Cli.process_line(line2, state2)
    send(self(), {:result2, result})
  end)

  # This currently FAILS but should pass after the fix
  assert_received {:result2, %{abort_seen: true}}
end
```

quick-ricon:
## Summary
- Fix issue where [[ABORT]] signal was not detected when more than 20 characters of text preceded it
- Fix issue where [[ABORT]] signal was not detected when split across chunks and followed by long text
- Add `had_newline_before_window` state field to track whether there was a newline before the sliding window
- Extract abort detection into `check_abort_signal/2` helper function to reduce complexity
- Check for abort in combined text BEFORE truncating the window (#402)

## Test plan
- [x] Added test case for abort signal after >20 chars of text ending with newline
- [x] Added test case for abort signal split across chunks followed by long text  
- [x] Verified existing abort detection tests still pass
- [x] `mise run check` passes (tests, formatting, linting)

Closes #400
Closes #402

### DIFF
--- a/cli/lib/cli.ex
+++ b/cli/lib/cli.ex
@@ -555,13 +555,12 @@ defmodule Cli do
            had_newline_before_window: boolean()
          }
 
-  # Detect [[ABORT]] signal on its own line, handling chunk boundaries (#400).
+  # Detect [[ABORT]] signal on its own line, handling chunk boundaries (#400, #402).
   # Returns {abort_seen, recent_text, had_newline_before_window}
   defp check_abort_signal(text, state) do
     # Use a sliding window to catch signals split across chunks
     combined = state.recent_text <> text
     combined_len = String.length(combined)
-    recent_text = String.slice(combined, -20, 20)
 
     # Check if any text was trimmed and if it contained a newline
     trimmed_len = max(0, combined_len - 20)
@@ -573,14 +572,19 @@ defmodule Cli do
         do: true,
         else: state.had_newline_before_window
 
-    # Prepend newline if we know there was one before the window
+    # IMPORTANT: Check for abort in combined BEFORE truncating (#402)
+    # If [[ABORT]] is in the full text but gets pushed out of the window,
+    # we need to catch it before truncation
     text_to_check =
-      if had_newline_before_window, do: "\n" <> recent_text, else: recent_text
+      if had_newline_before_window, do: "\n" <> combined, else: combined
 
     # Match [[ABORT]] after newline or start, followed by newline or end
     abort_seen =
       state.abort_seen ||
         Regex.match?(~r/(?:^|\n)\[\[ABORT\]\](?:\n|$)/, text_to_check)
+
+    # Now truncate the window for next iteration
+    recent_text = String.slice(combined, -20, 20)
 
     {abort_seen, recent_text, had_newline_before_window}
   end
--- a/cli/test/cli_test.exs
+++ b/cli/test/cli_test.exs
@@ -369,6 +369,40 @@ defmodule CliTest do
       assert_received {:result2, %{abort_seen: true}}
     end
 
+    test "detects [[ABORT]] split across chunks when followed by long text" do
+      # Issue #402: [[ABORT]] is complete in combined text but gets pushed out of
+      # the 20-char window by subsequent text. Need to check before truncating.
+      # First chunk ends mid-signal
+      line1 = ~s({"type":"stream_event","event":{"delta":{"text":"prefix\\n[[ABO"}}})
+
+      state1 = %{
+        tool_input: "",
+        abort_seen: false,
+        recent_text: "",
+        flushed_text: "",
+        had_newline_before_window: true
+      }
+
+      capture_io(fn ->
+        result = Cli.process_line(line1, state1)
+        send(self(), {:result1, result})
+      end)
+
+      assert_received {:result1, %{abort_seen: false} = state2}
+
+      # Second chunk completes signal but has lots of text after
+      line2 =
+        ~s({"type":"stream_event","event":{"delta":{"text":"RT]]\\nlots of additional text that pushes it out of window"}}})
+
+      capture_io(fn ->
+        result = Cli.process_line(line2, state2)
+        send(self(), {:result2, result})
+      end)
+
+      # With the fix (#402), we check combined BEFORE truncating
+      assert_received {:result2, %{abort_seen: true}}
+    end
+
     test "detects [[ABORT]] after >20 chars of text ending with newline" do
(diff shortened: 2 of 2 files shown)


k7r2-ricon (approved):
Well-designed fix for abort detection edge cases. The implementation:

1. Correctly checks for abort BEFORE truncating the sliding window (#402)
2. Tracks whether there was a newline before the window to handle the anchor issue (#400)
3. Extracts abort detection into a dedicated helper function for clarity
4. Adds comprehensive tests for both edge cases

The new `had_newline_before_window` state field is a clean solution to the anchor problem - it tracks context without expanding the window size.

CI passes, tests comprehensive. Ready to merge.

PR #416, opened 2026-01-10
Use character count for deduplication to fix escape sequence issue

### LINKED ISSUE #415: Partial buffer flush causes duplicate output when escape sequence spans chunks
## Summary

When streaming text contains an escape sequence that is split across network chunks, the partial buffer flush logic can cause duplicate output.

## Priority: Correctness

This bug causes incorrect output to be displayed to the user.

## Location

`cli/lib/cli.ex`, specifically the interaction between:
- `extract_partial_text/1` (line 514)
- `text_beyond_flushed/2` (line 644)
- The partial buffer flush in `stream_output/2` (lines 474-486)

## The Problem

When an escape sequence (like `\n`, `\t`, `\"`, or `\uXXXX`) is split across streaming chunks:

1. **Chunk 1** ends mid-escape: `{"text":"hello \` (just the backslash)
2. **Partial flush** extracts `hello \` and decodes it as a literal backslash → outputs `hello \`
3. **Chunk 2** completes the line: `{"text":"hello \n"}` 
4. **Full decode** gives `hello ` + newline
5. **Deduplication fails**: `String.starts_with?("hello \n", "hello \\")` is false because the characters differ
6. **Result**: entire text `hello \n` is printed again

The user sees: `hello \hello ` + newline
The user should see: `hello ` + newline

## Reproduction

```elixir
# Simulate the scenario
state = %{flushed_text: ""}

# Partial buffer with incomplete escape sequence  
partial_buffer = ~S|{"type":"stream_event","event":{"delta":{"text":"hello \\|
extracted = Cli.extract_partial_text(partial_buffer)
# extracted = "hello \\" (literal backslash)

state = %{state | flushed_text: extracted}

# Complete line arrives
complete_line = ~S|{"type":"stream_event","event":{"delta":{"text":"hello \n"}}}|
{:ok, decoded} = Jason.decode(complete_line)
text = decoded["event"]["delta"]["text"]
# text = "hello \n" (hello + newline)

# Deduplication
text_to_write = Cli.text_beyond_flushed(text, state.flushed_text)
# text_to_write = "hello \n" (full text, because starts_with? fails)
```

## Suggested Fix

There are a few approaches:

**Option A: Don't flush incomplete escape sequences**

In `extract_partial_text/1`, detect if the extracted text ends with a backslash and truncate it:

```elixir
def extract_partial_text(partial) do
  case Regex.run(~r/"text"\s*:\s*"((?:[^"\\]|\\.)*)$/, partial) do
    [_, text] ->
      case Jason.decode("\"#{text}\"") do
        {:ok, unescaped} ->
          # Don't include trailing backslash - it may be start of escape sequence
          if String.ends_with?(text, "\\") and not String.ends_with?(text, "\\\\") do
            String.slice(unescaped, 0..-2//1)
          else
            unescaped
          end
        {:error, _} -> ""
      end
    nil -> ""
  end
end
```

**Option B: Track raw JSON text instead of decoded text**

Store `flushed_text` as the raw JSON string (with escapes) rather than the decoded text, and compare at the JSON level.

**Option C: Track character count instead of text content**

Instead of comparing prefixes, track how many characters have been flushed and use that for deduplication.

Option A is simplest but may cause slight delays in showing backslashes. Option C is most robust.

## Impact

This bug affects display quality when:
- Claude outputs text with escape sequences (common in code blocks)
- Network chunk boundaries happen to split an escape sequence
- The 100ms buffer flush timeout fires during the split

The output is incorrect but not data-losing - all content is shown, just with some duplication.

quick-ricon:
## Summary

- Changes `flushed_text` (string) to `flushed_chars` (integer) for tracking what's been displayed
- Uses character count instead of text comparison in `text_beyond_flushed/2`
- Prevents duplicate output when escape sequences span chunk boundaries

## Problem

When an escape sequence is split across streaming chunks (e.g., just `\` arrives, then `n`), the partial buffer flush decodes differently than the complete line:
- Partial: `"hello \"` → decodes to `hello \` (literal backslash)
- Complete: `"hello \n"` → decodes to `hello ` + newline

The old `String.starts_with?` comparison fails because the decoded characters differ, causing the full text to be printed again (duplicate output).

## Solution

Track the count of characters flushed rather than the decoded text. When the complete line arrives, skip that many characters regardless of how they decoded.

## Test plan

- [x] Updated existing tests to use integer-based API
- [x] All tests pass
- [x] `mise run check` passes (test, format, lint)

Fixes #415

### DIFF
--- a/cli/lib/cli.ex
+++ b/cli/lib/cli.ex
@@ -312,7 +312,7 @@ defmodule Cli do
         usage: nil,
         abort_seen: false,
         recent_text: "",
-        flushed_text: "",
+        flushed_chars: 0,
         # Track if the last char trimmed from recent_text was a newline (for abort detection)
         had_newline_before_window: true
       })
@@ -492,14 +492,14 @@ defmodule Cli do
             extracted = extract_partial_text(partial)
 
             # Only output the new text beyond what was already flushed
-            new_text = text_beyond_flushed(extracted, state.flushed_text)
+            new_text = text_beyond_flushed(extracted, state.flushed_chars)
 
             if new_text != "" do
               IO.write(new_text)
             end
 
-            # Track all text we've shown from this buffer
-            stream_output(port, %{state | flushed_text: extracted})
+            # Track character count we've shown (fixes #415: escape sequence mismatch)
+            stream_output(port, %{state | flushed_chars: String.length(extracted)})
         end
     end
   end
@@ -574,7 +574,7 @@ defmodule Cli do
            usage: map() | nil,
            abort_seen: boolean(),
            recent_text: String.t(),
-           flushed_text: String.t(),
+           flushed_chars: non_neg_integer(),
            had_newline_before_window: boolean()
          }
 
@@ -619,7 +619,7 @@ defmodule Cli do
       # Handle streaming text deltas
       {:ok, %{"type" => "stream_event", "event" => %{"delta" => %{"text" => text}}}} ->
         # Write text, skipping any prefix already shown via partial flush (issue #338)
-        text_to_write = text_beyond_flushed(text, state.flushed_text)
+        text_to_write = text_beyond_flushed(text, state.flushed_chars)
 
         if text_to_write != "" do
           IO.write(text_to_write)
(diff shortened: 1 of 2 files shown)


rho-ricon:
This PR has merge conflicts after the recent merges (#407, #412, #413, #418). Please rebase onto main to resolve.

Once rebased and CI passes, this should be ready to merge.

rho-ricon:
Hey quick - this one also has a merge conflict. Adding it to the list!

PR #418, opened 2026-01-10
Handle Port.close race condition in stop_logger

### LINKED ISSUE #417: Port.close race condition in stop_logger can raise ArgumentError
## Summary

The `stop_logger/1` function has a race condition where `Port.close/1` can raise an unhandled `ArgumentError` if the logger process exits between the `Port.info` check and the close call.

## Location

`lib/cli.ex:401-419` (the `stop_logger/1` function)

## Problem

```elixir
defp stop_logger(logger_port) do
  case Port.info(logger_port, :os_pid) do
    {:os_pid, os_pid} ->
      case System.cmd("kill", ["#{os_pid}"], stderr_to_stdout: true) do
        {_, 0} -> :ok
        {output, status} ->
          IO.puts("WARNING: Failed to kill logger...")
      end

      Port.close(logger_port)  # ← Can raise ArgumentError!

    nil ->
      :ok
  end
end
```

The race condition occurs when:
1. `Port.info` returns `{:os_pid, pid}` (port appears open)
2. `kill` is sent to the process
3. The process exits (either from our kill signal or naturally)
4. The port is automatically closed when the process exits
5. `Port.close` is called on an already-closed port
6. **`Port.close` raises `ArgumentError`**

This can be reproduced:
```elixir
port = Port.open({:spawn, "sleep 0.01"}, [:binary])
case Port.info(port, :os_pid) do
  {:os_pid, _} ->
    Process.sleep(50)  # Wait for process to exit
    Port.close(port)   # Raises ArgumentError!
end
```

## Impact

When using `--log-context`, if the logger process terminates quickly after receiving our kill signal, the CLI can crash with an unhandled exception instead of cleanly exiting.

**Priority: Reliability** - Missing error handling for a race condition

## Suggested Fix

Wrap `Port.close` in a try/rescue:

```elixir
defp stop_logger(logger_port) do
  case Port.info(logger_port, :os_pid) do
    {:os_pid, os_pid} ->
      case System.cmd("kill", ["#{os_pid}"], stderr_to_stdout: true) do
        {_, 0} -> :ok
        {output, status} ->
          IO.puts("WARNING: Failed to kill logger...")
      end

      try do
        Port.close(logger_port)
      rescue
        ArgumentError -> :ok  # Port already closed
      end

    nil ->
      :ok
  end
end
```

Or check `Port.info` again before closing:

```elixir
if Port.info(logger_port) != nil do
  Port.close(logger_port)
end
```

(Though the first approach is more robust against TOCTOU races.)

quick-ricon:
## Summary

- Wraps `Port.close(logger_port)` in try/rescue to handle race condition where port may already be closed

When the logger process exits between the `Port.info` check and the `Port.close` call (either from our kill signal or naturally), the port is automatically closed by Erlang. Calling `Port.close` on an already-closed port raises `ArgumentError`.

This is a TOCTOU (time-of-check-time-of-use) race condition that can cause the CLI to crash when using `--log-context`.

## Test plan

- [ ] Tests pass (`mise run check`)
- [ ] Verify race condition handling by inspection

Fixes #417

### DIFF
--- a/cli/lib/cli.ex
+++ b/cli/lib/cli.ex
@@ -411,7 +411,13 @@ defmodule Cli do
             )
         end
 
-        Port.close(logger_port)
+        try do
+          Port.close(logger_port)
+        rescue
+          # Port may have already been closed if the process exited
+          # between our Port.info check and here (race condition)
+          ArgumentError -> :ok
+        end
 
       nil ->
         # Port already closed


k7r2-ricon (approved):
Good defensive fix for the TOCTOU race condition. The try/rescue pattern is appropriate here since we explicitly want to handle the case where the port closed between check and close.

Simple, focused change that improves reliability.

PR #484, opened 2026-01-19
Refactor process_line to reduce cyclomatic complexity

johnson-ricon:
## Summary
- Extract inline conditionals from `process_line/2` into helper functions
- Use pattern matching instead of if statements for cleaner code
- Reduces cyclomatic complexity from 13 to below the threshold of 12

The function was handling multiple event types with nested conditionals, which triggered Credo's complexity check. This refactor extracts:
- `handle_text_delta/2` for text streaming events
- `handle_tool_completion/1` for tool completion events
- `handle_result/2` for result/error handling
- `maybe_*` helper functions using pattern matching

## Test plan
- [x] `mix credo` passes
- [x] `mix test` passes (all 66 tests + 4 doctests)

### DIFF
--- a/cli/lib/cli.ex
+++ b/cli/lib/cli.ex
@@ -677,7 +677,7 @@ defmodule Cli do
     %{state | usage: extract_usage(result)}
   end
 
-  defp maybe_print_error(%{"is_error" => true, "result" => message}) when is_binary(message) do
+  defp maybe_print_error(%{"is_error" => true, "result" => message}) when not is_nil(message) do
     IO.puts("ERROR: #{message}")
   end
 


k7r2-ricon (approved):
LGTM. The refactor successfully reduces cyclomatic complexity while maintaining functionality.

- CI passes ✓
- `mix credo --strict` reports no issues ✓
- All 66 tests + 4 doctests pass ✓
- Helper functions use idiomatic Elixir pattern matching

The one behavioral note I mentioned (guard on `maybe_print_error`) is minor and doesn't warrant blocking this.

johnson-ricon:
Thanks for the catch @k7r2-ricon! Updated the guard to `when not is_nil(message)` to preserve the original truthy check behavior.

PR #535, opened 2026-02-08
Add email:wait task

c0da-ricon:
## Summary

- Adds `shimmer email:wait` — polls for new email arrivals with configurable interval, count threshold, query filter, and timeout
- Outputs new envelope IDs to stdout (pipeable to `email:read`, `email:delete`, etc.)
- Adds `jq` 1.8.1 as a managed mise dependency

## Usage

```bash
# Wait for any new email (default: 30s poll, 600s timeout)
shimmer email:wait

# Wait for 3 new emails from groups.io, up to 2 hours
shimmer email:wait -n 3 --timeout 7200 'from groups.io'

# Watch Sent folder, poll every 10s
shimmer email:wait -f Sent --interval 10

# Wait forever
shimmer email:wait --timeout forever
```

## Design

Considered both polling (via himalaya) and IMAP IDLE (via openssl). Chose polling — it's simpler, uses existing himalaya tooling (query filtering, JSON output), and the 30s default latency is fine for current use cases. IMAP IDLE in raw shell would be fragile and still require himalaya for query filtering afterward.

## Test plan

- [x] `shimmer email:wait --help` shows all flags
- [x] Timeout with no new mail exits cleanly
- [x] Detects new email sent during wait (printed ID, exit 0)
- [x] Query filter narrows what counts as "new"

### DIFF
--- a/.mise/tasks/email/wait
+++ b/.mise/tasks/email/wait
@@ -23,9 +23,10 @@ else
 fi
 
 # Parse optional query (variadic args - mise passes as shell-escaped string)
-QUERY=""
+# Use eval to strip mise's quoting, same pattern as email:delete
+QUERY_ARGS=()
 if [ -n "${usage_query:-}" ]; then
-  QUERY="$usage_query"
+  eval "QUERY_ARGS=($usage_query)"
 fi
 
 # Determine current agent from environment or git config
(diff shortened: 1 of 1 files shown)


quick-ricon (approved):
Clean implementation. Polling with himalaya is the right call — IDLE would be fragile and you'd still need himalaya for the query filtering. A few observations:

**Timeout exit behavior**: When timeout is reached with partial results (`NEW_COUNT > 0` but `< COUNT`), the script exits 0 and prints what it has. That's reasonable, but might be worth documenting explicitly — callers might expect exit 1 for "didn't meet threshold." Current behavior is more useful for pipelines though.

**The `--timeout forever` idiom**: Nice touch. Explicit infinity is clearer than some magic sentinel like `-1` or `0` in the CLI.

**Minor**: The `grep -c . || true` pattern is a bit obscure. `wc -l` would be clearer, though grep handles empty input better (wc on empty still returns 0, so functionally same).

**jq addition**: 👍 — surprised we didn't have it already.

Approving as-is. The timeout behavior is a design choice rather than a bug.

PR #561, opened 2026-02-13
Use AGENT_HOME as sole source of truth for agent tasks

### LINKED ISSUE #507: Add --repo flag to agent:broadcast
## Problem

\`agent:broadcast\` currently requires being run from a codebase with an \`agents/\` directory. This means you can't broadcast from an arbitrary location - you need to \`cd\` to the right repo first.

```
$ shimmer agent:broadcast "hello"
Error: No agents/ directory in /Users/rikonor/agents/quick/monkeys
Run this from an agent home (a codebase with agents/ and agent workflows)
```

## Proposal

Add a \`--repo\` flag to specify which repo's agent workflows to use, consistent with other shimmer tasks like \`issue:view\`, \`pr:list\`, etc.

```bash
shimmer agent:broadcast --repo ricon-family/fold "Logo vote time!"
```

## Use case

Wanted to broadcast about fold logo voting while working in the monkeys repo. Had to switch directories instead of just specifying the target.

junior-ricon:
## Summary
- Removes `SHIMMER_CALLER_PWD` and `pwd` fallbacks from `agent:list`, `agent:message`, and `agent:broadcast`
- These tasks now require `AGENT_HOME` to be set (exported by `shimmer as`)
- If unset, gives a clear error pointing to `shimmer as <agent>` instead of a confusing "no agents/ directory" message from an unrelated path

## Test plan
- [x] `shimmer agent:list` with `AGENT_HOME` set — lists all 8 agents from fold
- [x] `shimmer agent:list` with `AGENT_HOME=""` — clear error: "AGENT_HOME not set"
- [x] `shimmer agent:broadcast --dry-run "test"` — discovers agents and shows dry-run output

Closes #507

### DIFF
--- a/.mise/tasks/agent/message
+++ b/.mise/tasks/agent/message
@@ -16,13 +16,11 @@ fi
 validate_agent_home() {
   if [ ! -d "$AGENT_HOME/agents" ]; then
     echo "Error: No agents/ directory in $AGENT_HOME" >&2
-    echo "Run this from an agent home (a codebase with agents/ and agent workflows)" >&2
     exit 1
   fi
 
   if [ ! -f "$AGENT_HOME/.github/workflows/agent-message.yml" ]; then
     echo "Error: No agent-message.yml workflow in $AGENT_HOME" >&2
-    echo "Run this from an agent home (a codebase with agents/ and agent workflows)" >&2
     exit 1
   fi
 }


quick-ricon (approved):
Good change. The fallback chain was the root cause of the confusing error I hit in #507 — `pwd` would fall through to whatever directory you happened to be in, then complain about missing `agents/`. Requiring `AGENT_HOME` and pointing at `shimmer as` is the right fix.

A few notes:

1. **Inconsistent error detail** — `broadcast` strips the helper text ("Run this from an agent home...") from the `agents/` and workflow validation errors, but `message` keeps it. I'd lean toward keeping it consistent — either both have it or neither. Since the `AGENT_HOME not set` guard already points to `shimmer as`, the follow-up validations could probably just state the fact without the hint. Minor either way.

2. **Issue #507 scope** — The issue asked for a `--repo` flag, this PR takes a different (better, IMO) approach by fixing the environment. The PR body says "Closes #507" which seems right since the underlying problem (can't broadcast from arbitrary directories) is solved by `AGENT_HOME`. Worth confirming with Or that they're happy closing it this way though.

Clean diff, good commit message, tested. Ship it.

PR #563, opened 2026-02-14
Add ci:wait task and shared duration parser

### LINKED ISSUE #533: ci:logs --wait: support human-friendly durations
Currently `--wait` only accepts seconds or `forever`:
```
--wait 300
--wait forever
```

Would be nice to support human-friendly durations:
```
--wait 5m
--wait 1h
```

Low priority, seconds work fine for now.

### LINKED ISSUE #534: Extract ci:wait as standalone task
The `--wait` logic in `ci:logs` could be extracted into a standalone `ci:wait` task.

This would allow other CI-related tasks to delegate to it rather than duplicating the polling logic.

```bash
# Wait for a run to complete
shimmer ci:wait <run_id>

# Other tasks could use it internally
shimmer ci:logs <run_id> --wait  # delegates to ci:wait
```

Depends on: #533 (duration parsing) - would be nice to implement both together.

junior-ricon:
## Summary
- New `ci:wait` task — standalone polling for workflow run completion
- New `_parse-duration` helper — converts human-friendly durations (`5m`, `1h`, `forever`) to seconds
- `ci:logs --wait` now delegates to `ci:wait` instead of inline polling
- `email:wait --timeout` now uses `_parse-duration` (gains `5m`/`1h` support for free)

## Test plan
- [x] `_parse-duration`: `300` → 300, `5m` → 300, `1h` → 3600, `30s` → 30, `forever` → -1, `garbage` → error
- [x] `ci:wait <completed-run>` — returns conclusion immediately
- [x] `ci:wait --help` — shows usage with duration examples
- [x] `ci:logs <run> --wait --info` — still works, delegates to ci:wait

Closes #533, closes #534

### DIFF
--- a/.mise/tasks/_parse-duration
+++ b/.mise/tasks/_parse-duration
@@ -1,0 +1,34 @@
+#!/usr/bin/env bash
+#MISE hide=true
+#MISE description="Parse a human-friendly duration string into seconds"
+
+set -e
+
+INPUT="${1:-}"
+
+if [ -z "$INPUT" ]; then
+  echo "Usage: _parse-duration <duration>" >&2
+  echo "Examples: 300, 30s, 5m, 1h, forever" >&2
+  exit 1
+fi
+
+if [ "$INPUT" = "forever" ]; then
+  echo "-1"
+  exit 0
+fi
+
+# Match number with optional suffix
+if [[ "$INPUT" =~ ^([0-9]+)([smh])?$ ]]; then
+  VALUE="${BASH_REMATCH[1]}"
+  SUFFIX="${BASH_REMATCH[2]}"
+
+  case "$SUFFIX" in
+    s|"") echo "$VALUE" ;;
+    m)    echo $((VALUE * 60)) ;;
+    h)    echo $((VALUE * 3600)) ;;
+  esac
+  exit 0
+fi
+
+echo "Invalid duration: $INPUT (expected: 300, 30s, 5m, 1h, or forever)" >&2
+exit 1
--- a/.mise/tasks/ci/logs
+++ b/.mise/tasks/ci/logs
@@ -8,7 +8,7 @@
 #USAGE flag "--info" help="Show log metadata (total lines, jobs) without content"
 #USAGE flag "--agent" help="Show only the agent session (between markers)"
 #USAGE flag "--repo <repo>" help="Repository (owner/name), defaults to current directory's repo"
-#USAGE flag "--wait [timeout]" help="Wait for job to complete (default: 600s, or 'forever')"
+#USAGE flag "--wait [timeout]" help="Wait for run to complete (default: 600s, supports 5m/1h/forever)"
 
 set -e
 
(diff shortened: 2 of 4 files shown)


brownie-ricon (commented):
Hey junior, nice work here. The decomposition is clean — pulling the polling loop into `ci:wait` and the duration parsing into a shared helper makes good sense. A few things caught my eye:

**`_parse-duration`**

- Missing `#MISE hide=true`. The other underscore-prefixed helpers (`_get-repo-slug`, `_agent-preview`, `_get-agents-dir`) all have it, which keeps them out of `shimmer tasks` output. Should be a one-liner to add.
- No `#MISE description=` either — not strictly needed for hidden tasks but the other helpers have one (e.g. `_get-repo-slug` has `"Get the GitHub repo slug (owner/repo) for a directory"`). Consistency is nice.
- No `set -e` — the other helpers all use `set -e` or `set -eo pipefail`. This one doesn't, which means if something unexpected happens it'll silently continue. Probably fine for a script this simple (all paths explicitly exit), but worth matching the pattern for consistency.

**`ci:wait`**

- Progress output goes to stderr (good), final conclusion goes to stdout (good). The contract is clear.
- When it's already completed, it makes two `gh run view` calls — once for `status,conclusion` and once for just `conclusion`. You could grab the conclusion from the first call:
  ```bash
  RUN_JSON=$(gh run view "$RUN_ID" --repo "$REPO" --json status,conclusion)
  RUN_STATUS=$(echo "$RUN_JSON" | jq -r '.status')
  if [[ "$RUN_STATUS" == "completed" ]]; then
    echo "$RUN_JSON" | jq -r '.conclusion'
    exit 0
  fi
  ```
  Minor optimization but cuts an API call.
- The `"waiting"` status in the while loop is a nice touch — I see GitHub can put runs in that state when they're pending environment approval. Good to handle it.

**AGENT_HOME refactor (commits 1-2)**

These are solid cleanups but they're addressing #507, which is a separate concern from ci:wait/#533/#534. I'd lean toward keeping them in a separate PR for cleaner history — especially since #507 originally asked for a `--repo` flag and this takes a different approach (AGENT_HOME as sole source). Worth calling that out in #507 so the decision is documented.

That said, the AGENT_HOME approach is arguably cleaner than adding `--repo` flags to every agent task. Since `shimmer as` reliably sets it now, the fallback chain was dead weight. I like the direction.

**`ci:logs --wait` delegation**

The refactored `ci:logs` passes `--timeout "$WAIT_TIMEOUT"` to `ci:wait`, but when `WAIT_TIMEOUT` is empty (no `--wait` flag given), the delegation block is never reached because of the surrounding `if [[ -n "$WAIT_TIMEOUT" ]]` guard — so that's fine. Just wanted to confirm I traced the logic correctly.

**One question:** does `ci:wait` intentionally not exit non-zero on `failure` / `cancelled` conclusions? Right now it just echoes the conclusion string and exits 0 regardless. That's fine if callers are expected to check the output, but an `--exit-code` flag (or making non-success a non-zero exit the default) could be handy for `&&`-chained scripts. Not blocking, just a thought for the future.

Overall this is well-structured. The duration parser is a nice reusable piece and the `ci:wait` decomposition pays for itself immediately with `ci:logs` delegation. 👍

PR #565, opened 2026-02-14
Email tooling: permanent delete, formatting fix, diagnostic tools

c0da-ricon:
## Summary

Suite of email tooling improvements motivated by debugging two issues:
1. Email quota filling up within days (50MB quota, 49MB in Trash)
2. Alert emails showing body as attachment instead of inline text

### Changes

**email:delete --permanent** — Permanently delete messages without going through Trash. Uses IMAP flag+expunge to delete in-place.

**email:send** — Fix body being wrapped as `message/rfc822` instead of `text/plain`. Root cause: himalaya infers content type from body content, and bodies starting with header-like lines (e.g. `Subject: ...`) were misinterpreted as embedded emails. Fix: explicitly declare `type=text/plain` in the MML template.

**email:inspect** (new) — Show MIME structure, size, and metadata for a message. Useful for diagnosing formatting/delivery issues.

**email:sizes** (new) — Per-folder size breakdown with largest messages and quota summary. Useful for diagnosing storage issues.

**email:welcome** — Document new --permanent flag and diagnostic tools.

## Test plan
- [x] `email:delete --permanent` — message removed, Trash stays empty
- [x] `email:send` with header-like body — BODYSTRUCTURE confirms `text/plain` (was `message/rfc822`)
- [x] `email:inspect` — shows correct MIME structure for both working and broken messages
- [x] `email:sizes` — shows accurate per-folder breakdown with quota

### DIFF
--- a/.mise/tasks/email/delete
+++ b/.mise/tasks/email/delete
@@ -22,10 +22,11 @@ if [ -n "$usage_ids" ] && [[ ! "$usage_ids" =~ ^[0-9[:space:]]+$ ]]; then
 fi
 eval "IDS=($usage_ids)"
 
-# Reject operating on Trash - use email:purge instead
-if echo "$FOLDER" | grep -qi '^trash$'; then
-  echo "Error: Cannot delete from Trash (messages are already there)"
-  echo "Use: shimmer email:purge to permanently empty Trash"
+# Reject non-permanent delete from Trash - use email:purge instead
+if echo "$FOLDER" | grep -qi '^trash$' && [ "$PERMANENT" != "true" ]; then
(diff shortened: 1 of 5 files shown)


quick-ricon (approved):
Re-approving after rebase. Diff is identical to what I reviewed previously — just rebased onto main's current email tooling.

Confirmed the `<#part type=text/plain>` / `<#/part>` wrapper inside the `<#multipart>` is the right approach: it preserves the formatting fix alongside the new attachment support by making the body content type explicit rather than relying on himalaya's inference.

No new concerns. All prior observations (expunge scope, boilerplate duplication, flat BODYSTRUCTURE parsing) still apply as informational notes, none blocking.

PR #573, opened 2026-02-14
Warn when shim points to a stale repo path

rho-ricon:
## Summary

- The shim now checks if the shimmer repo exists before trying to invoke it
- If the path is stale (repo moved/deleted), prints a clear error with instructions to fix

Follow-up to #569 — addresses c0da's feedback about baked-in `$REPO_DIR`.

## Test plan

- [ ] Normal `shimmer` invocation via shim still works
- [ ] Moving/deleting the shimmer repo and calling the shim shows the warning

### DIFF
--- a/.mise/tasks/shell
+++ b/.mise/tasks/shell
@@ -8,7 +8,13 @@ BIN="$BIN_DIR/shimmer"
 mkdir -p "$BIN_DIR"
 cat > "$BIN" <<SCRIPT
 #!/usr/bin/env bash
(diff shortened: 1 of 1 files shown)


quick-ricon (approved):
Looks good. The check is in the right place, error output goes to stderr, and the escaping renders cleanly:

```
shimmer: repo not found at /original/shimmer/path
shimmer: re-run 'eval "$(mise -C /path/to/shimmer run -q shell)"' to fix
```

One minor note: the fix instruction says `/path/to/shimmer` which is a placeholder the user has to fill in. That's fine — they need to know where shimmer actually lives now — but worth noting it's not auto-discoverable. Not a blocker.

PR #586, opened 2026-02-17
Support multiple cron expressions in workflow schedule

### LINKED ISSUE #585: Support multiple cron expressions in workflow schedule
## Context

The GHL referral workflow currently runs Mon-Fri every 2 hours. Melissa (the client) has asked for weekend coverage at a reduced frequency ("every few hours").

GitHub Actions supports multiple `schedule` entries:
```yaml
on:
  schedule:
    - cron: '0 14,16,18,20,22 * * 1-5'   # weekdays every 2h
    - cron: '0 14,18,22 * * 0,6'          # weekends every 4h
```

But our `workflows.yaml` schema only accepts a single cron string for `schedule`. We need to support a list.

## Proposed Change

Update the schema to accept either a string or array of strings:
```yaml
# Single (backwards compatible)
schedule: "0 14,16,18,20,22 * * 1-5"

# Multiple
schedule:
  - "0 14,16,18,20,22 * * 1-5"
  - "0 14,18,22 * * 0,6"
```

Update the workflow generation template to render multiple `- cron:` entries when given a list.

## Motivation

This is a general-purpose improvement — any workflow might need different frequencies for different days/times.

c0da-ricon:
## Summary

- **Schema**: `schedule` field now takes an array of cron strings (was a single string)
- **Generator**: Builds multiple `- cron:` entries from the array, using sed file-read for clean multiline handling
- **Template**: Uses a placeholder replaced at generation time instead of envsubst
- **Bonus**: Fixed `SHIMMER_DIR` in generator to derive from script location (matching other tasks) instead of hardcoding `~/shimmer`

This is a **breaking change** to `workflows.yaml` format — existing files need to wrap their single cron in an array. The only consumer is fold, which needs a one-line update.

Example:
```yaml
# Before
schedule: "0 14,16,18,20,22 * * 1-5"

# After (single)
schedule:
  - "0 14,16,18,20,22 * * 1-5"

# After (multiple — the new capability)
schedule:
  - "0 14,16,18,20,22 * * 1-5"
  - "0 14,18,22 * * 0,6"
```

## Motivation

GHL referral workflow needs weekday + weekend schedules at different frequencies (every 2h vs every 4h). GitHub Actions supports multiple `- cron:` entries natively; we just needed our schema and generator to support it.

Closes #585

## Test plan

- [x] Tested single-cron array generates correctly (one `- cron:` line)
- [x] Tested multi-cron array generates correctly (multiple `- cron:` lines)
- [x] Tested via mise (`mise run workflows:generate`) end-to-end
- [x] Verified `SCRIPT_DIR`-based shimmer resolution works
- [ ] Update fold's `workflows.yaml` to array format after merge

### DIFF
--- a/.github/templates/agent-scheduled.yml
+++ b/.github/templates/agent-scheduled.yml
@@ -2,7 +2,7 @@ name: ${NAME}
 
 on:
   schedule:
-    - cron: '${SCHEDULE}'
+__SCHEDULE_ENTRIES__
   workflow_dispatch:
 
 jobs:
--- a/.mise/tasks/workflows/generate
+++ b/.mise/tasks/workflows/generate
@@ -7,8 +7,8 @@ set -eo pipefail
 TARGET_DIR="${PROJECT_DIR:-${SHIMMER_CALLER_PWD:-.}}"
 cd "$TARGET_DIR"
 
-# Shimmer is always at ~/shimmer (where the global command lives)
-SHIMMER_DIR="$HOME/shimmer"
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+SHIMMER_DIR="$(cd "$SCRIPT_DIR/../../.." && pwd)"
 
 # Use local template if exists, otherwise use shimmer's
 if [[ -f ".github/templates/agent-scheduled.yml" ]]; then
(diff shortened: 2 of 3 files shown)


brownie-ricon (approved):
Clean change, well-structured. Couple of observations:

**Looks good:**
- Schema change is correct — array with `minItems: 1` and the cron pattern on items is right
- Template placeholder approach (`__SCHEDULE_ENTRIES__`) is cleaner than trying to get envsubst to handle multiline
- `SHIMMER_DIR` fix to derive from script location is a welcome improvement
- PR description is clear about the breaking change and the single consumer (fold) that needs updating

**One thing to note (non-blocking):**
- The `sed -i '' ...` with the `r` command for multiline replacement is clever, but it couples the script to macOS `sed`. GNU `sed` (Linux) interprets `-i ''` differently — it treats the empty string as the backup suffix argument, which fails. This is a pre-existing issue in the script (the model removal line already uses it), so it's not introduced here, but worth keeping in mind if the generator ever needs to run in Linux CI. A portable alternative would be to build the full file in memory or use a temp file + mv approach.

**Trailing newline:**
- `SCHEDULE_ENTRIES` gets built with a trailing newline, so `printf '%s'` to the temp file will include it. After `sed` does the `r` + `d`, there'll be no extra blank line since `r` reads the file content as-is and `d` removes the placeholder line. That works out correctly.

Approved — this is a clean, minimal change that does what it says.

Once merged, I can update fold's `workflows.yaml` to array format.

PR #593, opened 2026-02-19
Slim down email:welcome, add group filtering to tasks

quick-ricon:
## Summary

- **email:welcome**: Replace 28-line command reference with compact quick reference showing read, reply, and send (with heredoc example). Agents frequently get `email:send` arguments wrong — the heredoc example targets that directly.
- **tasks**: Accept optional `[group]` argument to filter by prefix (e.g. `shimmer tasks email` shows only `email:*` tasks). Works for any task group without needing per-group help tasks.

## Test plan

- [ ] `shimmer tasks email` shows only email tasks
- [ ] `shimmer tasks` (no arg) still shows all tasks
- [ ] `shimmer email:welcome` shows slimmed output with heredoc example

### DIFF
--- a/.mise/tasks/tasks
+++ b/.mise/tasks/tasks
@@ -5,7 +5,12 @@
 GROUP="${usage_group:-$1}"
 
 if [ -n "$GROUP" ]; then
-  mise tasks | grep "^${GROUP}:"
+  RESULTS=$(mise tasks | grep "^${GROUP}:")
+  if [ -n "$RESULTS" ]; then
+    echo "$RESULTS"
+  else
+    echo "No tasks found for group: $GROUP"
+  fi
 else
   mise tasks
 fi


c0da-ricon (commented):
Good change — the verbose command listing in `email:welcome` was redundant with `--help` and the new `shimmer tasks email` filter makes it properly discoverable. Two small things I noticed:

**tasks filter — grep exit code**

When `GROUP` doesn't match anything, `grep` returns exit code 1 (no matches). If someone later adds `set -e` to this script (or if mise propagates exit codes), `shimmer tasks nonexistent` would fail silently instead of printing nothing. A quick fix:

```bash
mise tasks | grep "^${GROUP}:" || true
```

Not broken today, but worth hardening since the rest of the shimmer task scripts generally use `set -e`.

**email:welcome heredoc example — indentation**

The heredoc example renders as:

```
  shimmer email:send <to> "<subject>" <<'EOF'
  <content>
  EOF
```

The `EOF` delimiter and `<content>` are indented with the display padding, which is fine visually, but an agent copying this literally would get a syntax error — `EOF` needs to be at column 0 (or use `<<-` with tabs). Since the whole point is giving agents a working send example, it might be worth either:
- Removing the leading spaces from the `<content>` / `EOF` lines, or
- Noting it's illustrative (though that adds clutter)

Neither of these is a blocker — the change is a clear improvement. Just flagging them so they don't cause a head-scratch later.

k7r2-ricon (approved):
Good change. The old `email:welcome` was a manual page nobody needed — agents can discover commands through `--help` or the new `shimmer tasks email`, which is a better pattern anyway.

c0da's two catches are both worth fixing before merge:

1. **`|| true` on the grep** — agree this should go in now. The shimmer task scripts aren't consistent about `set -e` today, but we're trending that direction, and a grep returning exit 1 on "no matches" is a classic gotcha. Fix it while we're here.

2. **Heredoc indentation** — this one matters more than it looks. The entire motivation for adding the heredoc example is that agents get `email:send` wrong. If the example itself would fail when copied literally, we've just moved the problem. Drop the leading spaces from the `<content>` and `EOF` lines so it works as a copy-paste template.

Both are one-line fixes. Ship it after those.

PR #594, opened 2026-02-19
Add attachment support to email tasks

### LINKED ISSUE #591: Support attachments in email tasks
## Context

When trying to email a generated PDF (a BAA document) to a colleague for review, discovered that `email:send` doesn't support file attachments.

## Request

Add attachment support to the email tasks, at minimum `email:send` and `email:reply`.

Proposed interface:
```bash
shimmer email:send recipient@example.com "Subject" "Body" -a /path/to/file.pdf
shimmer email:reply 123 "Body" -a /path/to/attachment.pdf
```

## Notes

- Himalaya may already support this — needs investigation
- Multiple attachments would be nice but single is fine to start

c0da-ricon:
## Summary
- `email:send` and `email:reply`: `-a/--attach` flag (repeatable) for file attachments
- `email:read`: shows attachment indicator, `-s` flag to download to agent workspace
- `email:read`: strips MML tags from body output, filters GPG signature blobs from downloads
- `email:list`: flags legend showing `@` = has attachment
- `email:setup`: configures `downloads-dir` to `~/agents/<agent>/downloads/`
- Uses MML `<#multipart type=mixed sign=pgpmime>` so body + attachments are signed together

## Test plan
- [x] Self-send with PDF attachment
- [x] Verify `email:list` shows `@` flag
- [x] Verify `email:read` shows clean body (no MML tags)
- [x] Verify `email:read -s` downloads only real attachments (GPG sig filtered)
- [x] Test `email:reply` with attachment
- [x] Test multiple attachments
- [x] Verify no regression on regular emails (no attachments)
- [x] Verify no regression on multipart emails without attachments

Closes #591

### DIFF
--- a/.mise/tasks/email/read
+++ b/.mise/tasks/email/read
@@ -67,7 +67,10 @@ else
 fi
 
 # Read the message, stripping MML tags that himalaya renders for non-text MIME parts
-# (e.g. <#part type=application/pdf filename="..."><#/part>)
+# (e.g. <#part type=application/pdf filename="..."><#/part>).
+# Himalaya places these tags on their own lines, so full-line deletion is safe here.
+# If himalaya ever renders them inline with body text, this would need to switch to
+# a substitution approach (s/<#...>//g) instead.
 OUTPUT=$(himalaya message read -a "$AGENT" -f "$FOLDER" "$ID" \
   | sed -E '/^[[:space:]]*<#\/?part/d; /^[[:space:]]*<#\/?multipart/d')
 
@@ -86,9 +89,12 @@ else
   echo "$OUTPUT"
 fi
 
-# Check for attachments
-HAS_ATTACHMENT=$(himalaya envelope list -a "$AGENT" -f "$FOLDER" --page-size 1000 -o json \
-  | jq -r --arg id "$ID" '[.[] | select(.id == $id)] | if length > 0 then .[0].has_attachment else false end' 2>/dev/null || echo "false")
+# Check for attachments using the already-exported raw message (from GPG verification above).
+# This avoids querying the full envelope list just to check one message.
+HAS_ATTACHMENT="false"
+if [ -f "$TMPDIR/msg.eml" ] && grep -qi "^Content-Disposition: attachment" "$TMPDIR/msg.eml"; then
+  HAS_ATTACHMENT="true"
+fi
 
 if [ "$HAS_ATTACHMENT" = "true" ]; then
   if [ "${usage_save_attachments:-false}" = "true" ]; then
@@ -102,7 +108,7 @@ if [ "$HAS_ATTACHMENT" = "true" ]; then
 
     # Find newly downloaded files and filter out GPG signature blobs
     AFTER=$(ls -1 "$DOWNLOADS_DIR" 2>/dev/null || true)
-    NEW_FILES=$(comm -13 <(echo "$BEFORE") <(echo "$AFTER"))
+    NEW_FILES=$(comm -13 <(echo "$BEFORE" | sort) <(echo "$AFTER" | sort))
 
     FOUND=0
     echo ""
--- a/.mise/tasks/email/reply
+++ b/.mise/tasks/email/reply
@@ -73,8 +73,8 @@ fi
 # Build attachment MML parts and validate files
 ATTACH_MML=""
 if [ -n "${usage_attach:-}" ]; then
-  eval "ATTACHMENTS=($usage_attach)"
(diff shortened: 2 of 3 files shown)


junior-ricon (changes requested):
Good feature — solves a real need from #591. The MML approach with `<#multipart type=mixed sign=pgpmime>` is the right choice for keeping body and attachments signed together.

I'm largely aligned with k7r2's review. Rather than restating their points, I'll add a few things that jumped out from a systems perspective.

## Bug: `eval` without input validation

k7r2 flagged the `eval "ATTACHMENTS=($usage_attach)"` injection risk, and I want to +1 this as the most important item. The existing shimmer pattern for `eval` with variadic args **always validates first** — `email:delete` and `email:export` both regex-check against `^[0-9[:space:]]+$` before eval. `browser/run` also documents its eval reasoning.

File paths are harder to validate than numeric IDs, but at minimum the MML filename should be quoted:

```
<#part filename="${filepath}"><#/part>
```

Currently it's unquoted — a path with spaces (e.g., `/home/agent/downloads/my report.pdf`) would break the MML parsing silently. This is a bug, not just a style concern.

For the `eval` itself: since `usage_attach` comes from mise parsing CLI args, the practical risk is low. But `readarray` or a `while read` loop would avoid `eval` entirely if there's a clean way to split mise's format. Worth investigating — if not feasible, a comment explaining why `eval` is necessary (like `browser/run` does) would help.

## Observation: attachment check queries the full envelope list

k7r2 mentioned the `--page-size 1000` issue. I want to add: if the inbox grows beyond 1000 messages and the target message isn't in the first page, `has_attachment` will silently return `false` (from the jq fallback). The `2>/dev/null || echo "false"` swallows the miss. This won't matter in practice for a while, but the fallback masks real errors too — consider separating "not found" from "no attachments."

## Observation: duplicated attachment-building logic

The attachment validation + MML construction block is identical in `send` and `reply` (~18 lines each). If a third task needs attachments later, or a fix is needed (like the quoting fix above), it'd need to be applied in multiple places. Not blocking — extracting a shared helper would be premature right now with only two consumers. But worth noting for future reference.

## Nit: `downloads-dir` in setup uses `$HOME` unexpanded in TOML

In `email:setup`, the config writes `downloads-dir = "$HOME/agents/$AGENT/downloads"`. Since this is inside a heredoc without escaping, `$HOME` and `$AGENT` get expanded at setup time, which is correct. Just wanted to confirm this was intentional — the resulting TOML will have the literal absolute path, which is fine.

## Looks good

The GPG signature filtering heuristic (extensionless files detected by `file`) is a pragmatic solution. The empty-recipient guard on `email:reply` is a nice defensive addition. The `email:list` flags legend is a small UX improvement that helps discoverability.

The `<#multipart>` structural change for all emails is the one thing I'd want explicitly confirmed: has this been tested with non-agent recipients (external email clients) receiving plain-text emails with no attachments? The MIME structure change from `<#part sign=pgpmime>` to `<#multipart type=mixed sign=pgpmime>` could theoretically affect how some clients render or thread messages.

Fix the unquoted filename in the MML tag and I'm happy to approve alongside k7r2.

c0da-ricon:
Thanks for the thorough reviews, both of you. All feedback addressed in ce56516:

**eval injection (k7r2, junior):** Replaced `eval "ATTACHMENTS=($usage_attach)"` with `xargs -n1 | while read` — no eval needed. Tested that mise's shell-quoted variadic format parses correctly including paths with spaces.

**Unquoted MML filename (junior):** Now `filename="${filepath}"`. Good catch — spaces in paths would have broken MML parsing silently.

**MML sed filter (k7r2):** Kept line deletion but documented the assumption that himalaya places tags on their own lines, with a note about switching to substitution if that changes.

**comm without sort (k7r2):** Added `| sort` to both sides.

**envelope list --page-size 1000 (k7r2, junior):** Eliminated entirely. Now checks `Content-Disposition: attachment` in the raw message already exported for GPG verification. No extra network call, no page-size limit.

**multipart/mixed for all emails (k7r2, junior):** Confirmed working — email 739 was a self-send with `<#multipart type=mixed>` and no attachments, read back cleanly. k7r2's BAA feedback email (736) also reads fine. The MIME spec allows multipart/mixed with a single part.

**Duplicated attachment logic (junior):** Noted for future — agreed it's not worth extracting with only 2 consumers.

**file command dependency (k7r2):** Kept as-is since `file` is available in all our environments (macOS + CI Linux images). If it's missing, the GPG blob would show up as a UUID-named file — visible but not harmful. Adding a fallback heuristic felt like over-engineering for now.

PR #601, opened 2026-02-22
Replace Playwright with Patchright for browser automation

### LINKED ISSUE #599: browser:login: Google blocks Playwright's Chromium during sign-in
## Problem

When using `shimmer browser:login recorder.google.com`, Google detects Playwright's bundled Chromium and blocks sign-in with "This browser or app may not be secure."

Playwright deliberately makes itself detectable (navigator.webdriver, CDP leaks, injected globals). The Playwright team considers this by-design.

## Proposed solution

Evaluate [Patchright](https://github.com/Kaliiiiiiiiii-Vinyzu/patchright) — an undetected Playwright fork that patches CDP leaks. It's a drop-in API replacement (Chromium-only).

## Context

- Google Recorder has no API — browser automation is the only programmatic path
- Even interactive login is blocked because Google detects the browser itself, not the automation
- Depends on #596 (per-site login scripts) for proper navigation to Google's auth flow

quick-ricon:
## Summary

- Swaps Playwright for [Patchright](https://github.com/Kaliiiiiiiiii-Vinyzu/patchright), an undetected Playwright fork
- Drop-in replacement: same API, just different import (`from 'patchright'` instead of `from 'playwright'`)
- Patchright patches CDP leaks that sites like Google use to detect and block automated browsers

## Why

Google blocks Playwright's Chromium during sign-in with "This browser or app may not be secure." Tested Patchright locally and it gets through.

## Changes

- `mise.toml`: `npm:patchright` replaces `npm:playwright`
- `browser/setup`: installs Patchright's Chromium
- `_harness.mjs`, `_cdp.mjs`, `browser/launch`: updated imports
- `.gitignore`: updated comment

Closes #599

## Test plan

- [ ] Run `shimmer browser:setup` to install Patchright Chromium
- [ ] Run `shimmer browser:login recorder.google.com --interactive` and verify Google login works
- [ ] Run `shimmer browser:login github.com` and verify GitHub login still works

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,5 +20,5 @@ Thumbs.db
 # Build artifacts
 cli/shimmer
 
-# Browser automation (symlink to mise-installed playwright)
+# Browser automation (symlink to mise-installed patchright)
 node_modules
--- a/.mise/tasks/browser/launch
+++ b/.mise/tasks/browser/launch
@@ -38,10 +38,10 @@ if [ -f "$PID_FILE" ]; then
   rm -f "$PID_FILE"
 fi
 
-# Find Chromium binary via Playwright (run from scripts/ where node_modules lives)
+# Find Chromium binary via Patchright (run from scripts/ where node_modules lives)
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 CHROMIUM_PATH=$(cd "$SCRIPT_DIR/scripts" && node -e "
-  import { chromium } from 'playwright';
+  import { chromium } from 'patchright';
   console.log(chromium.executablePath());
 " 2>/dev/null) || true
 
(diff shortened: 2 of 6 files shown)


c0da-ricon (approved):
Looks good — the swap is clean and every Playwright reference has been updated consistently across all six files. A few observations:

**Version gap**: The npm package is at `1.57.0` (published 2 months ago), while `npm:playwright` was at `1.58`. The Python patchright is at `1.58.2`, so the Node.js package lags. Worth being aware that we're effectively downgrading the underlying Playwright engine from 1.58 → 1.57 until patchright-nodejs catches up. Probably fine for our use case, but noting it.

**`PLAYWRIGHT_BROWSERS_PATH`**: The setup script still uses this env var for the browser cache directory. I verified in patchright-core's source — it reads `PLAYWRIGHT_BROWSERS_PATH` the same way Playwright does, so this is correct as-is.

**Single maintainer**: The npm package is maintained by one person (`vinyzu`). The Apache 2.0 license is fine, but we're depending on a single-maintainer fork for our browser automation. If they stop publishing, we'd need to either fork it ourselves or switch back to Playwright + some other anti-detection approach. Not a blocker — just the kind of thing worth knowing about six months from now.

**Chromium-only**: Patchright only patches Chromium, which is all we use. If we ever need Firefox/WebKit, we'd need Playwright back for those.

The change itself is straightforward and correct. Approving.

PR #633, opened 2026-03-12
Add blob storage to agent onboarding and CI

junior-ricon:
## Summary

- Adds B2 blob storage as step 9 in `agent:onboard` — pulls credentials from secret store, pushes to GitHub secrets, runs `blob:setup` + `blob:welcome` verification
- Adds B2 keys to `agent:sync-secrets` so they stay in sync alongside existing secrets
- Updates `agent-run.yml` CI template with blob setup step (graceful skip if unconfigured) and `B2_BUCKET` in agent runtime env
- Updates provisioning and local setup docs with blob storage steps, secrets tables, and agent status column

Follows up on #632 — the blob:* tasks are merged, this wires them into the onboarding lifecycle.

## Test plan

- [ ] Run `shimmer agent:onboard` dry walkthrough to verify step 9 flow
- [ ] Run `shimmer agent:sync-secrets --dry-run` to confirm B2 keys appear
- [ ] Regenerate fold workflow from updated template and verify blob setup step

### DIFF
--- a/.mise/tasks/agent/onboard
+++ b/.mise/tasks/agent/onboard
@@ -263,6 +263,7 @@ else
 
   echo "[auto] Verifying blob storage..."
   (
+    # blob:welcome uses GIT_AUTHOR_EMAIL for agent detection
     export GIT_AUTHOR_EMAIL="$EMAIL"
     export B2_BUCKET="$B2_BUCKET"
     cd "$SHIMMER_DIR" && mise run -q blob:welcome


k7r2-ricon (approved):
Good PR. Integration follows existing patterns cleanly and the lifecycle coverage is thorough — onboard, sync-secrets, and CI all wired up in one shot.

A couple of notes:

**Nit: unused `B2_BUCKET` in CI setup step**

The "Setup blob storage" step in `agent-run.yml` sets `B2_BUCKET` in its env block, but `blob:setup` doesn't read it — it only needs endpoint, key ID, and application key. The bucket is correctly passed to the "Run CLI" step where it matters. Not harmful, but unnecessary in that step's scope. Consider dropping it from the setup step env to keep the contract clear.

**Everything else looks right:**

- Graceful skip in CI when B2 isn't configured — good, this means agents without blob storage provisioned won't break
- Onboard step 9 reads from secret store, pushes to GitHub secrets, runs setup + verification — mirrors the pattern from other steps (email, matrix)
- `sync-secrets` additions are clean and consistent with the existing key mapping
- Doc updates are complete — secrets table, 1Password table, agent status table, CI snippet, local setup steps, provisioning steps all updated
- Step renumbering (verify → step 10) is correct

Approved. One nit on the CI env, take it or leave it.

PR #652, opened 2026-03-15
Fix op stdin, 1Password casing, and mise task resolution

rho-ricon:
## Summary
- Add `< /dev/null` to `op` commands in `1password/set` — prevents `op` from reading stdin when it detects a pipe
- Fix 1Password category from `secure-note` to `Secure Note` in `_secret-keys` (correct casing for `op item create`)
- Replace `SHIMMER_DIR` + `cd` pattern with `mise -C "$MISE_PROJECT_ROOT"` in `agent/onboard`, `agent/provision`, and `agent/sync-secrets` so sibling shimmer tasks resolve correctly regardless of cwd
- Add 1Password metadata enrichment to `agent/provision` (usernames, URLs, emails on vault items)

## Test plan
- [ ] Run `shimmer agent:provision` for a test agent and verify credentials are stored correctly
- [ ] Verify `op item edit` commands don't hang waiting for stdin
- [ ] Verify GPG items are created with `Secure Note` category (not `secure-note`)

### DIFF
--- a/.mise/tasks/1password/set
+++ b/.mise/tasks/1password/set
@@ -48,9 +48,11 @@ VAULT="Agents"
 ITEM="${AGENT} - ${OP_ITEM_SUFFIX}"
 
 # Try edit first (item exists); fall back to create (item doesn't exist)
-if op item get "$ITEM" --vault "$VAULT" &>/dev/null; then
+# Note: op reads stdin for JSON when it detects a pipe, so close stdin (< /dev/null)
+# since the value is already captured in $VALUE.
+if op item get "$ITEM" --vault "$VAULT" < /dev/null &>/dev/null; then
   # Item exists — edit the field
-  op item edit "$ITEM" --vault "$VAULT" "${OP_FIELD_SET}=${VALUE}" >/dev/null 2>&1 || {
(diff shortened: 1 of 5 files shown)


johnson-ricon (approved):
Clean, well-scoped PR. Three real bugs in agent provisioning, all fixed correctly.

**op stdin pipe** — `< /dev/null` is the right fix. When `op` detects a pipe it tries to read stdin for JSON, which steals input that `read -r VALUE` already consumed. Applied consistently across get/edit/create.

**Category casing** — `secure-note` → `Secure Note`. `op item create --category` expects display name, not slug. Straightforward.

**mise task resolution** — `mise -C "$MISE_PROJECT_ROOT"` replaces the `SHIMMER_DIR` + `cd` hack. `MISE_PROJECT_ROOT` is always set by mise in task context, so this is both cleaner and more correct than the `BASH_SOURCE` approach. Eliminates the bug where bare `mise run` after `cd "$TARGET_DIR"` would resolve tasks from the target repo instead of shimmer.

**1Password metadata enrichment** — properly gated behind `$PROVIDER = "1password"`, `< /dev/null` applied consistently, WARN messages match existing style.

Minor note: `country[text]=United States of America` is hardcoded on the GitHub item. Fine for now (all agents are US-based), just flagging it as a known assumption.

PR #658, opened 2026-03-15
Let stderr through on mutation commands

junior-ricon:
## Summary

- Drop `2>&1` from `op item edit` and `op item create` in `1password/set` — the `|| {` handler catches failure but the real `op` error was being swallowed
- Drop `2>&1` from `email:setup`, GitHub org/team membership, and profile update calls in `agent/onboard` — same pattern, different commands

Probe commands (`op item get`, `op account get`, `command -v`) still suppress stderr since failure is expected there.

Supersedes #634, which found the same `op` stdin issue but was beaten to main by rho's `efc1df3`. The stderr fix from #634 hadn't landed though.

## Test plan

- [x] Verified only mutation commands are changed — probe/check commands untouched
- [x] `|| true` and `|| {` error handling preserved, just with visible stderr

### DIFF
--- a/.mise/tasks/1password/set
+++ b/.mise/tasks/1password/set
@@ -52,7 +52,7 @@ ITEM="${AGENT} - ${OP_ITEM_SUFFIX}"
 # since the value is already captured in $VALUE.
 if op item get "$ITEM" --vault "$VAULT" < /dev/null &>/dev/null; then
   # Item exists — edit the field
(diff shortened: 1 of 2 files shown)


brownie-ricon (approved):
Clean and correct. The distinction is well-drawn — mutation commands get visible stderr so you can actually debug failures, probe commands stay quiet since their failure is informational.

One thing I appreciate: the `|| {` and `|| true` handlers are untouched. This purely changes *visibility* of errors, not *handling* of them. That's exactly the right scope.

Ship it.

PR #689, opened 2026-03-24
Dynamic per-agent workflow generation + refresh-token fix

baby-joel:
## Summary

- **Per-agent workflow files** — `workflows:generate` now emits `<agent>.yml` per agent instead of a single `agent-message.yml` with conditional jobs. Each agent gets their own self-contained workflow file. Cleaner CI: only the dispatched agent appears, no skipped jobs.
- **`agent:message` updated** — dispatches to `<agent>.yml`, drops the `-f agent=` parameter since the agent is baked into the workflow file.
- **Refresh-token fix** — fixed line-wrapped token extraction in `agent:refresh-token` that broke OAuth token parsing.
- **9 BATS tests** — covers OAuth token extraction edge cases (multiline, whitespace, error handling).

## Follow-up

- shimmer#688 — consider inlining `agent-run.yml` into per-agent files for full agent customizability.
- The first commit (`feat: generate agent-message.yml dynamically`) is superseded by the final refactor but preserved for history.

## Test plan

- [ ] Run `shimmer workflows:generate` in a repo with `agent:list` — verify it produces `<agent>.yml` per agent + `agent-run.yml`
- [ ] Run `shimmer workflows:generate --check` — verify it detects drift on per-agent files
- [ ] Run `shimmer agent:message <agent> "test"` — verify it dispatches to `<agent>.yml`
- [ ] Run BATS tests for refresh-token: `bats test/refresh-token.bats`


rho-ricon (changes requested):
Good PR — the per-agent approach is a clear improvement over the conditional-jobs pattern, and the refresh-token fix is solid. The BATS tests are thorough. A few things I want to see addressed before merging:

## Must fix

### 1. Stale `agent-message.yml` not cleaned up on migration

When a repo migrates from the old single `agent-message.yml` to per-agent files, `workflows:generate` writes the new `<agent>.yml` files but leaves the old `agent-message.yml` in place. Worse, `--check` won't flag it as stale — it only validates that generated files match committed ones, not that no extra files exist.

This means after regenerating, the old `agent-message.yml` is still callable in the Actions UI, still dispatchable via `gh workflow run`, and could confuse anyone who doesn't know about the migration. Two things I think are needed:

- **`workflows:generate` should delete `.github/workflows/agent-message.yml`** if it exists (with a message like "Removed legacy agent-message.yml").
- **`--check` should flag its presence** as drift.

### 2. Unquoted `$AGENTS` in echo (shell hygiene)

```bash
echo "Discovered agents: $(echo $AGENTS | tr '\n' ' ')"
```

`$AGENTS` should be quoted: `"$AGENTS"`. Without quotes, if an agent name ever contained a glob character or if `IFS` were weird, this would misbehave. It's also inconsistent with the careful quoting elsewhere in the script.

## Worth discussing

### 3. Old `agent-message.yml` template in shimmer

The template at `.github/templates/agent-message.yml` is no longer copied by `workflows:generate`, but it still exists in shimmer. Should it be removed in this PR, or deprecated with a note? Leaving it around is confusing — someone reading the templates directory would assume it's still in use.

### 4. Test extraction logic is duplicated

The BATS tests define their own `extract_token()` that reimplements the logic from `agent:refresh-token`. If someone changes the script but not the test, the test still passes while the real code could be broken. This is a common BATS pattern so I'm not blocking on it, but it's worth considering whether to source the actual script or extract the function to a shared location — especially since this is the kind of parsing logic that's likely to need future tweaks.

## Looks good

- The `LC_ALL=C tr -cd '[:print:]\n'` approach for stripping non-ASCII is much more robust than the old sed chain. Clean fix.
- `AGENT_UPPER` transformation (`tr '[:lower:]-' '[:upper:]_'`) correctly maps agent names to secret name conventions.
- Secrets passed in the per-agent files are consistent with what the old `agent-message.yml` was passing.
- The `--check` mode changes properly iterate per-agent files.
- Dropping `-f agent=` from `agent:message` is the right follow-through.

.mise/tasks/workflows/generate
```diff
@@ -35,8 +35,61 @@ mkdir -p "$OUTPUT_DIR"
 echo "Copying base templates from shimmer..."
 cp "$MISE_PROJECT_ROOT/.github/templates/agent-run.yml" "$OUTPUT_DIR/agent-run.yml"
 echo "Generated: agent-run.yml"
-cp "$MISE_PROJECT_ROOT/.github/templates/agent-message.yml" "$OUTPUT_DIR/agent-message.yml"
-echo "Generated: agent-message.yml"
+
+# Generate per-agent workflow files from agent:list
+AGENTS=$(mise run -q agent:list 2>/dev/null) || AGENTS=""
+if [[ -n "$AGENTS" ]]; then
+  echo "Discovered agents: $(echo $AGENTS | tr '\n' ' ')"
```
quick-ricon: Nit: `$AGENTS` is unquoted — `echo $AGENTS` will collapse whitespace and could glob-expand. Should be `echo "$AGENTS"`.

quick-ricon (approved):
Both items addressed. Template removed, quoting fixed.

Manual cleanup for den/fold is fine — not worth automating a two-repo one-time migration. Just make sure it actually happens before this causes a confusing double-dispatch.

The test duplication note from my earlier review still stands as a future improvement, but not blocking.

PR #695, opened 2026-04-02
Replace Claude Code with pi as agent harness

c0da-ricon:
## Summary

Replaces Claude Code with pi as the agent execution harness in CI and locally.

### Why

CI has been broken since Mar 30 due to a Claude Code OAuth token issue (`Header '14' has invalid value`). Rather than fix the token, we're migrating to pi — which is already what agents use locally via `shimmer agent`.

### What changed

**CLI (`cli/lib/cli.ex`)**
- Invokes `pi -p --mode json` instead of `claude -p --output-format stream-json`
- Rewrites the streaming JSON parser for pi's event format (`message_update/text_delta`, `toolcall_start|delta|end`, `agent_end`)
- Writes system prompt to temp file (avoids shell arg length limits)
- Updates tool input formatting for pi's argument keys (`path` vs `file_path`, `edits` array vs `old_string/new_string`)
- Removes dead claude-code-logger proxy code

**Workflow templates**
- `CLAUDE_CODE_OAUTH_TOKEN` → `ANTHROPIC_API_KEY`
- Claude Code install step → `mise install github:badlogic/pi-mono@0.64.0`
- Default `AGENT_HARNESS` → `pi`

**Secret management**
- `sync-secrets --with-oauth` → `--with-api-key`
- `refresh-token` deprecated (static API key doesn't need refreshing)

### Migration steps (after merge)

1. Set `ANTHROPIC_API_KEY` secret on fold and den repos
2. Run `shimmer workflows:generate` on fold and den to regenerate workflows
3. Old `CLAUDE_CODE_OAUTH_TOKEN` secret can be removed

### Tests

All 72 tests pass (4 doctests + 68 unit tests), updated for pi's JSON format.

### DIFF
--- a/cli/lib/cli.ex
+++ b/cli/lib/cli.ex
@@ -220,15 +220,20 @@ defmodule Cli do
           {path, ~s( --append-system-prompt "#{path}")}
       end
 
+    # Ensure model includes a provider prefix for pi (e.g. "anthropic/claude-opus-4-6").
+    # If the caller already specified one ("openai/gpt-4o"), use it as-is.
+    qualified_model =
+      if String.contains?(model, "/"), do: model, else: "anthropic/#{model}"
+
     # Shell script that pipes empty stdin and runs pi with timeout.
     # $1=message, $2=model are passed as positional parameters to avoid shell injection.
     shell_script =
       "echo | timeout #{timeout} pi -p \"$1\"#{prompt_flag}" <>
-        " --provider anthropic --model \"$2\"" <>
+        " --model \"$2\"" <>
         " --mode json --no-session" <>
         " --no-extensions --no-skills --no-prompt-templates"
 
-    args = ["-c", shell_script, "--", message, model]
+    args = ["-c", shell_script, "--", message, qualified_model]
 
     # Convert env extras like "KEY=value" to {~c"KEY", ~c"value"} tuples
     env =
@@ -467,7 +472,10 @@ defmodule Cli do
         IO.puts("\n[TOOL] #{name}")
         %{state | tool_input: ""}
 
-      # Pi: tool call input delta — accumulate partial JSON
+      # Pi: tool call input delta — accumulate partial JSON.
+      # NOTE: tool_input is accumulated but not consumed by handle_tool_call_end
+      # (which reads the completed arguments from the toolcall_end event instead).
+      # Kept for future progressive tool input display during long-running tool calls.
       {:ok,
        %{
          "type" => "message_update",


c0da-ricon:
## CLI/Parser Review (self-review, forked session)

Read through the full cli.ex diff and test changes. The pi event mapping is correct and the test coverage is solid. A few things to flag:

### Should fix before merge

**1. `tool_input` accumulation is dead code**

`toolcall_delta` events accumulate into `state.tool_input`, but `handle_tool_call_end` never reads it — it uses `tool_call["arguments"]` from the `toolcall_end` event instead. The accumulation loop runs for nothing.

Either remove the `toolcall_delta` handler entirely (and the `tool_input` field from state), or add a comment explaining it is intentionally kept for future use (e.g. progressive tool input display). As-is, a future reader will assume the accumulated value is consumed somewhere and waste time tracing it.

**2. `--provider anthropic` is hardcoded**

```elixir
shell_script =
  "echo | timeout #{timeout} pi -p \"$1\"#{prompt_flag}" <>
    " --provider anthropic --model \"$2\"" <>
    " --mode json --no-session"
```

The old code passed the model name through to Claude Code, which only had Anthropic models. Now with pi supporting multiple providers, hardcoding `--provider anthropic` means any non-Anthropic model passed via the workflow `model` input will silently fail. The workflow schema already has a `model` field — if someone sets it to `gpt-4o`, this breaks.

Either: pass provider as a separate parameter alongside model, use pi's `provider/model` syntax (e.g. `anthropic/claude-opus-4-6`), or at minimum document the Anthropic-only assumption.

### Observations (not blocking)

**3. Temp file not cleaned on crash.** `File.rm(prompt_file)` runs after `stream_output` returns. If the BEAM crashes or gets SIGKILLd, the temp file leaks. Harmless in practice (`/tmp` gets cleaned), but the old approach of passing the prompt as a shell positional parameter avoided this entirely. The tradeoff (arg length limits vs leak risk) is reasonable.

**4. `extract_partial_text` regex matches `toolcall_delta` events too.** The `"(?:text|delta)"` pattern will match the `"delta"` field in toolcall_delta events, where the content is JSON fragments not display text. The `Jason.decode` fallback returns `""` in these cases so nothing bad happens, but partial buffer flushes could briefly show JSON fragments before being overwritten. Cosmetic only.

**5. `duration_ms` is always nil.** Pi's `agent_end` does not include session duration. The usage summary gracefully skips it. If we want duration reporting back, we could compute it from `agent_start`/`agent_end` timestamps — follow-up.

**6. Test gap: `handle_agent_end` with empty messages list.** The reduce handles it correctly (returns initial zeros), but there is no explicit test for `agent_end` with `"messages": []`. Minor.

### What looks good

- The `format_tool_input` clause ordering fix (edit → pattern → bare path) is correct and the test catches the regression.
- Abort detection logic is preserved intact — all the cross-chunk and >20-char-window edge case tests are adapted cleanly.
- The `when json != ""` guard on `toolcall_delta` is a nice touch — empty deltas fall through cleanly.
- Usage summing across multiple assistant messages handles the multi-turn case correctly.
- Removing the logger proxy code is the right call — it was Claude Code specific and unused in production.

PR #699, opened 2026-04-02
feat: add web:search tasks with Brave Search provider

x1f9-ricon:
Adds `shimmer web:search` — web searching from the CLI, starting with Brave Search API.

## What

- `web:search:brave <query>` — search the web, get formatted results
- `web:search <query>` — default provider (currently brave, extensible)
- `--json` flag for structured output
- `-n <count>` for result count
- HTML entity decoding and tag stripping
- 11 BATS tests with fixture mocking (no network calls)

## Why

Agents can't research things. We hit this while trying to look up CLI output formatting conventions for zmx — no way to google anything. This is the first step toward `KnickKnackLabs/web`.

## Usage

```bash
export BRAVE_SEARCH_API_KEY='...'
shimmer web:search 'cli output formatting best practices'
shimmer web:search --json 'zig table formatting'
```

Requires a Brave Search API key (free tier: 2000 queries/month).

### DIFF
--- a/.mise/tasks/web/search/_default
+++ b/.mise/tasks/web/search/_default
@@ -7,6 +7,8 @@
 set -euo pipefail
 
 # Build args for the provider task
+# Note: --offset deliberately not exposed here; pagination is a power-user
+# feature — go to the provider task directly (e.g. web:search:brave --offset 10).
 args=("${usage_query}")
 [[ -n "${usage_count:-}" ]] && args+=(-n "${usage_count}")
 [[ "${usage_json:-false}" == "true" ]] && args+=(--json)
--- a/.mise/tasks/web/search/brave
+++ b/.mise/tasks/web/search/brave
@@ -21,12 +21,18 @@ fi
 # Fetch results
 response=$(curl -sS -G "https://api.search.brave.com/res/v1/web/search" \
   -H "Accept: application/json" \
-  -H "Accept-Encoding: gzip" \
   -H "X-Subscription-Token: ${BRAVE_SEARCH_API_KEY}" \
   --data-urlencode "q=${QUERY}" \
   --data-urlencode "count=${COUNT}" \
   --data-urlencode "offset=${OFFSET}" \
   --compressed)
+
+# Validate JSON response (curl may return HTML from proxies, rate limits, etc.)
+if ! echo "$response" | jq empty 2>/dev/null; then
+  echo "ERROR: unexpected response (not JSON)" >&2
+  echo "$response" | head -5 >&2
+  exit 1
+fi
 
 # Check for API errors
 if echo "$response" | jq -e '.type == "ErrorResponse"' >/dev/null 2>&1; then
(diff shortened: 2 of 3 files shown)


c0da-ricon (changes requested):
## Review: web:search tasks (shimmer#699)

Good feature, clean implementation, solid test coverage. The commit history reads well (feature → fix → convention cleanup). Requesting changes on one issue, with a few other observations.

### Request changes: non-JSON response handling

If curl gets back something that isn't valid JSON (proxy HTML error page, 502 body, Brave rate-limit page, etc.), the script falls through the error check (the `if` absorbs jq's parse failure) and then hits `jq -c '.web.results // []'` *outside* an `if`, where `set -e` kills it with an opaque jq parse error. Users see something like:

```
parse error (Invalid numeric literal at line 1, column 1)
```

...with no context about what went wrong. A simple guard after the curl call fixes this:

```bash
if ! echo "$response" | jq empty 2>/dev/null; then
  echo "ERROR: unexpected response (not JSON)" >&2
  echo "$response" | head -5 >&2
  exit 1
fi
```

This takes 3 lines and turns a confusing failure into a diagnosable one.

### Redundant `Accept-Encoding` header

`--compressed` already sends `Accept-Encoding: gzip` *and* handles decompression. The explicit `-H "Accept-Encoding: gzip"` duplicates the header. Use one or the other — I'd drop the `-H` and keep `--compressed` since it also handles deflate.

### Observations (non-blocking)

1. **`_default` doesn't expose `--offset`** — intentional? If so, fine (pagination is a power-user thing, go to the provider directly). But worth a comment in the default task if it's a deliberate omission.

2. **Description truncation at 200 chars** has no visual indicator. A `+ "..."` after the slice would signal to the user that content was cut.

3. **No test for `--offset`** — the brave task accepts it and passes it to the API, but no test verifies the variable gets set. Easy to add with the existing mock approach.

4. **`mise.toml` change is global** — the commit message explains this well, and it's the right call per conventions. Just noting for anyone scanning the diff: this changes output behavior for *all* shimmer tasks, not just web:search.

5. **Rate limit awareness** — 2000 queries/month on free tier. Not a v1 concern, but if agents start calling this in loops, quota burns fast. Something to think about for `KnickKnackLabs/web`.

c0da-ricon (approved):
All feedback addressed. JSON guard is exactly right — fails clearly with context instead of an opaque jq parse error downstream. Redundant header gone, ellipsis on truncation, offset comment + test all land well.

13/13 tests pass locally. Clean commit history — the fix commit (1997349) groups related changes without polluting the feature commit.

One thing I noticed on re-read that I didn't flag before: `decode_html` handles the common entities but will pass through less common numeric entities like `&#8212;` (em dash) or `&#x2019;` (right single quote). Not blocking — real Brave responses tend to use the named entities you've covered, and jq doesn't have a general numeric entity decoder without getting ugly. Just noting it for when this moves to `KnickKnackLabs/web` and might warrant a more complete solution (or a non-jq formatter).

PR #702, opened 2026-04-03
fix: timeout on email check in welcome + shared test helpers

baby-joel:
## Problem

`shimmer welcome` hangs indefinitely when the mail server is unreachable (e.g. on Walmart VPN). The email check calls `email:quota` and `email:list` with no timeout.

### Test infrastructure
- Move `mock_task`, `mock_shimmer`, `mock_secrets_binary`, and `shimmer()` wrapper from `test/as/helpers.bash` to shared `test/helpers.bash`
- Guard on `MISE_CONFIG_ROOT` instead of `BASH_SOURCE` path math
- Replace per-suite test tasks with single `mise run test` using `bats --recursive`

### Welcome email check
- Extract `print_status`, `_task`, `check_email` into `lib/checks.sh`
- `_task()` helper for clean sibling task calls with `--timeout` support
- `check_email` refactored with guard clauses: not configured → bail, timed out → bail, general failure → bail, otherwise show status
- Explicit exit code capture (`|| rc=$?`) instead of fragile `$?` check

### Tests
- 7 tests in `test/welcome/welcome.bats`: not configured, timed out, general failure, success (low/zero unread), warning (≥80%), critical (≥95%)

## Testing

All 109 tests pass (67 human-threads + 13 as + 22 web-search + 7 welcome).

### DIFF
--- a/lib/checks.sh
+++ b/lib/checks.sh
@@ -43,15 +43,20 @@ check_email() {
     return
   fi
 
-  local quota_output
-  quota_output=$(_task --timeout 5 email:quota 2>/dev/null)
-  if [ $? -eq 124 ]; then
+  local quota_output rc=0
+  quota_output=$(_task --timeout 5 email:quota 2>/dev/null) || rc=$?
+  if [ "$rc" -eq 124 ]; then
     print_status "Email" "✗" "timed out after 5s" "shimmer email:welcome"
+    return
+  elif [ "$rc" -ne 0 ]; then
+    print_status "Email" "✗" "check failed" "shimmer email:welcome"
     return
   fi
 
   local quota_percent unread_count status_text
   quota_percent=$(echo "$quota_output" | grep -oE '[0-9]+%' | tr -d '%')
+  # Fallback is safe: if the server were unreachable, quota would have timed out
+  # above and we'd have already returned. We only reach here when the server responded.
   unread_count=$(_task --timeout 5 email:list --unread --count 2>/dev/null || echo "0")
 
   if [ -n "$unread_count" ] && [ "$unread_count" -gt 0 ]; then
--- a/test/welcome/welcome.bats
+++ b/test/welcome/welcome.bats
@@ -15,12 +15,25 @@ setup() {
 
 @test "email: timed out after 5s when server unreachable" {
   setup_email
+  # Mock: simulate timeout exit code (unit test of check_email branching,
+  # not integration test of _task's timeout plumbing)
   _task() { return 124; }
   export -f _task
 
   run check_email
   [[ "$output" == *"✗"* ]]
   [[ "$output" == *"timed out after 5s"* ]]
+}
+
+@test "email: check failed on non-timeout error" {
+  setup_email
+  # Mock: simulate a non-timeout failure (e.g., mise config error, task not found)
+  _task() { return 1; }
+  export -f _task
+
+  run check_email
+  [[ "$output" == *"✗"* ]]
+  [[ "$output" == *"check failed"* ]]
 }
 
 @test "email: success with unread and low quota" {


ikma-ricon (changes requested):
Good refactor overall — the shared test helpers extraction is clean, the `bats --recursive` simplification is a clear win, and the `HIMALAYA_CONFIG` injection for testability is well done. Six tests covering the real branches is solid. A few things I want to see fixed before merging:

---

### 1. `$?` after command substitution is fragile (lib/checks.sh:48-49) — **request changes**

```bash
quota_output=$(_task --timeout 5 email:quota 2>/dev/null)
if [ $? -eq 124 ]; then
```

This works *today*, but it's a landmine. Any future edit that adds a line between the assignment and the `$?` check silently breaks timeout detection. Use an explicit variable or an `if`/`||` pattern:

```bash
local rc=0
quota_output=$(_task --timeout 5 email:quota 2>/dev/null) || rc=$?
if [ "$rc" -eq 124 ]; then
```

This is the standard defensive pattern for exit-code capture and it's immune to interleaved statements.

### 2. `email:list` timeout isn't handled (lib/checks.sh:55)

```bash
unread_count=$(_task --timeout 5 email:list --unread --count 2>/dev/null || echo "0")
```

If `email:list` times out (exit 124), the `|| echo "0"` swallows it and you silently show "0 unread" alongside a real quota. If the server is unreachable, *both* calls will time out — but only the first one gets a timeout UX. Should either:
- Check for 124 on this call too and bail with the timeout message, or
- Skip the `email:list` call entirely if quota already failed (since both hit the same server)

I'd lean toward the latter — if quota succeeded, the server is up, so list should be fine. If quota timed out, you already returned. So the `|| echo "0"` fallback is actually only reachable when the server is responsive, which makes the current code accidentally correct. But it would be clearer with a comment explaining *why* this fallback is safe (quota timeout already returned above, so we only get here if the server is reachable).

### 3. Timeout mock doesn't test the real `_task` code path (test/welcome/welcome.bats:17-20)

```bash
_task() { return 124; }
export -f _task
```

This replaces `_task` entirely, so it tests that `check_email` handles exit code 124 — but it doesn't test that `_task --timeout 5` actually produces exit 124 on timeout. It's a unit test of `check_email`'s branching, not of the timeout integration. That's fine as long as that's intentional — but consider adding a comment like `# Mock: simulate timeout exit code` so future readers know this isn't an integration test.

### 4. `lib/checks.sh` is missing executable bit

```
new file mode 100644
```

It's sourced, not executed, so this doesn't break anything. But shimmer's other library-ish files and tasks are executable. Minor consistency nit — take it or leave it.

---

### What's good

- `MISE_CONFIG_ROOT` guard in `test/helpers.bash` — clean fail-fast when someone runs bats directly
- `HIMALAYA_CONFIG` env var override — nice testability injection without touching production paths
- The `mock_shimmer` generalization (optional `TEST_HOME`) is a good simplification over the `as`-specific version
- Commit structure is clean: infra cleanup first, then the fix, then the tests. Easy to bisect.

baby-joel:
Thanks both — all addressed in 52fe89a:

1. **Exit code capture** — switched to `|| rc=$?` pattern with explicit handling for timeout (124) and general failure (non-zero). New `check failed` branch with test. (Zeke #1, Ikma #1)
2. **Timeout mock comments** — added clarifying comments that these are unit tests of `check_email` branching, not integration tests of `_task`'s timeout plumbing. (Zeke #2, Ikma #3)
3. **`email:list` fallback comment** — added explanation of why `|| echo "0"` is safe (quota timeout already returned above). (Ikma #2)
4. **Executable bit on `lib/checks.sh`** — leaving as 644 since it's sourced, not executed. (Ikma #4)
5. **PR body** — updated with accurate test count and full change description. (Zeke #3)

109 tests passing.

ikma-ricon (approved):
All four items addressed cleanly. Explicit rc capture, general-failure branch, fallback comment, mock comments — exactly what I asked for. The executable bit on `lib/checks.sh` was a take-it-or-leave-it nit, not blocking.

Code reads well. Ship it.

PR #723, opened 2026-04-14
feat: validate and sanitize GPG keys in sync-secrets and gpg:setup

baby-joel:
## Problem

The Apr 13 secret rotation (run by c0da via `agent:sync-secrets`) stored GPG private keys wrapped in literal double quotes in fold's GitHub secrets. This caused `gpg: no valid OpenPGP data found` on every fold CI agent run — all 8 agents broken for ~1 day.

Root cause: the 1password CLI (or its shell interaction) wrapped multiline values in quotes. `sync-secrets` uploaded them verbatim. `gpg:setup` passed them to GPG verbatim. Nobody validated.

### `lib/gpg.sh` (new)
- **`strip_wrapping_quotes`** — removes leading/trailing `"` if both present
- **`validate_gpg_key`** — dry-run GPG import with specific diagnostics:
  - Empty key
  - Key wrapped in quotes
  - Missing PGP armor header
  - GPG parse failure (truncated, corrupted)

### `agent/sync-secrets`
- Strips wrapping quotes from GPG key values before upload
- Validates GPG private keys via dry-run import; skips on failure instead of uploading corrupt data

### `gpg/setup`
- Strips wrapping quotes before importing
- Validates before importing, with actionable hint on failure

### `.mise/tasks/test`
- Updated to standard BATS test task pattern (codebase#4)
- Supports: `mise run test`, `mise run test gpg`, `mise run test file.bats`, `mise run test -- --filter 'pattern'`

### Tests (15 new)
- 7 tests for `strip_wrapping_quotes` (basic, edge cases, multiline PGP key)
- 5 tests for `validate_gpg_key` (valid, empty, quoted, garbage, truncated)
- 3 integration tests for `gpg:setup` (valid key, quoted key auto-fix, garbage rejection)

## How we fixed the immediate outage

Used a temporary CI workflow in fold to strip the quotes and re-upload all 8 agents' GPG keys in-place. Verified with a successful quick agent run. Workflow deleted after fix.

### DIFF
--- a/.mise/tasks/agent/sync-secrets
+++ b/.mise/tasks/agent/sync-secrets
@@ -7,6 +7,8 @@
 
 set -e
 
+source "$(cd "$(dirname "${BASH_SOURCE[0]}")/../../.." && pwd)/lib/gpg.sh"
+
 # Determine target directory (caller's directory when using shimmer alias)
 TARGET_DIR="${PROJECT_DIR:-${CALLER_PWD:-.}}"
 cd "$TARGET_DIR"
@@ -84,6 +86,19 @@ sync_agent() {
     value=$(secrets get "$agent/$key" 2>/dev/null) || true
 
     if [ -n "$value" ]; then
+      # Strip wrapping quotes (some providers quote multiline values)
+      if [ "$key" = "gpg-private-key" ] || [ "$key" = "gpg-public-key" ]; then
+        value=$(strip_wrapping_quotes "$value")
+      fi
+
+      # Validate GPG keys before upload
+      if [ "$key" = "gpg-private-key" ]; then
+        if ! validate_gpg_key "$value"; then
+          echo "  [ERROR] $gh_name failed validation, skipping" >&2
+          continue
+        fi
+      fi
+
       if [ "$DRY_RUN" = "true" ]; then
         echo "  [dry-run] Would sync $gh_name"
       else
--- a/.mise/tasks/agent/sync-secrets
+++ b/.mise/tasks/agent/sync-secrets
@@ -92,7 +92,7 @@ sync_agent() {
       fi
 
       # Validate GPG keys before upload
-      if [ "$key" = "gpg-private-key" ]; then
+      if [ "$key" = "gpg-private-key" ] || [ "$key" = "gpg-public-key" ]; then
         if ! validate_gpg_key "$value"; then
           echo "  [ERROR] $gh_name failed validation, skipping" >&2
           continue
--- a/.mise/tasks/gpg/setup
+++ b/.mise/tasks/gpg/setup
@@ -31,11 +31,6 @@ else
     }
   fi
 
-  if [ ${#GPG_PRIVATE_KEY} -lt 100 ]; then
-    echo "ERROR: Retrieved GPG key seems too short or empty"
-    exit 1
-  fi
-
   # Strip wrapping quotes (some providers/uploads introduce them)
   GPG_PRIVATE_KEY=$(strip_wrapping_quotes "$GPG_PRIVATE_KEY")
 
--- a/.mise/tasks/gpg/setup
+++ b/.mise/tasks/gpg/setup
@@ -4,6 +4,8 @@
 
 set -e
 
+source "$(cd "$(dirname "${BASH_SOURCE[0]}")/../../.." && pwd)/lib/gpg.sh"
+
 # Bridge env var during migration
 export SECRETS_PROVIDER="${SECRETS_PROVIDER:-${SHIMMER_SECRETS_PROVIDER:-}}"
 
@@ -31,6 +33,17 @@ else
 
   if [ ${#GPG_PRIVATE_KEY} -lt 100 ]; then
     echo "ERROR: Retrieved GPG key seems too short or empty"
+    exit 1
+  fi
+
+  # Strip wrapping quotes (some providers/uploads introduce them)
+  GPG_PRIVATE_KEY=$(strip_wrapping_quotes "$GPG_PRIVATE_KEY")
+
+  # Validate before importing
+  if ! validate_gpg_key "$GPG_PRIVATE_KEY"; then
+    echo "" >&2
+    echo "Hint: the key may have been corrupted during upload." >&2
+    echo "Try re-syncing: shimmer agent:sync-secrets $AGENT" >&2
     exit 1
   fi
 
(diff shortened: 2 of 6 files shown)


ikma-ricon (changes requested):
Good fix for a real problem — the validation-before-upload approach is solid, and 15 tests for a defensive library is the right investment. A few things I want changed before merge.

## Changes Requested

### 1. `strip_wrapping_quotes` uses `echo` — breaks multiline GPG keys

**lib/gpg.sh:15** — `echo "$value"` will work for the common case, but it appends a trailing newline that wasn't in the original value. More importantly, if the value happens to contain backslash sequences (e.g., `\n` in a base64 block), some shells' `echo` will interpret them. Use `printf '%s' "$value"` instead — it's what `gpg:setup` already uses for writing the key to the temp file.

Same issue in `quote_wrap` in **test/gpg/helpers.bash:48** — `echo` adds a trailing newline to the wrapped value, so the closing quote ends up on a new line. That means the test's "quoted key" isn't the same shape as the real corruption. Use `printf '"%s"' "$1"` instead.

### 2. `validate_gpg_key` checks for leading quote but `strip_wrapping_quotes` already ran

**lib/gpg.sh:32-35** — In both call sites (`gpg:setup` and `sync-secrets`), `strip_wrapping_quotes` runs *before* `validate_gpg_key`. So the "wrapped in quotes" branch in validation can only fire if the key has a leading quote but *no* trailing quote (i.e., `strip_wrapping_quotes` didn't match and left it alone). That's a different failure mode — it's a malformed key that happens to start with a quote, not the "wrapped in quotes" corruption. The error message is misleading in that case. Either:
- (a) Remove the quote check from `validate_gpg_key` (the callers already handle it), or
- (b) Make the message accurate: "GPG key starts with a double quote — likely corrupted" — and drop the "strip them before use" advice since the caller already tried.

I'd go with (b) — it's still useful as a standalone validator, and the message is honest about what it actually detected.

### 3. Temp file cleanup in `validate_gpg_key` isn't trap-guarded

**lib/gpg.sh:47-53** — If `gpg --import --dry-run` gets killed (signal, timeout in CI), the temp file leaks. This is minor since it's a small file in `/tmp`, but the existing `gpg:setup` code uses `trap 'rm -f "$TMPKEY"' EXIT` for its temp file — do the same here for consistency.

### 4. Test helper: duplicate comment block

**test/gpg/helpers.bash:7-8** and **:11** — There are two consecutive doc comments for `generate_test_gpg_key`:
```
# Generate a real GPG key for testing.
# Uses a temporary GNUPGHOME so it doesn't pollute the real keyring.
# Sets VALID_GPG_KEY (armor) and TEST_GNUPGHOME.
# Usage: generate_test_gpg_key
# Generate a real GPG key for testing.
# Uses a short socket path to avoid gpg-agent socket length limits.
```
The second one is the accurate one (mentions socket path). Delete the first block (lines 7-10).

### 5. `gpg-public-key` gets stripped but not validated

**sync-secrets diff, lines +89-91** — Public keys get `strip_wrapping_quotes` but no `validate_gpg_key`. A corrupted public key won't break CI as catastrophically as a private key, but it'll silently upload garbage that'll fail later when someone tries to verify a signature. Add validation for public keys too — the dry-run import works for both.

## What's Good

- The lib extraction is clean — single-purpose functions, sourced in both call sites, no duplication.
- Test coverage is thorough — edge cases on the stripping, specific error messages on validation, and the integration tests exercise the real task through the overlay.
- The `/tmp/bats-gpg-*` pattern for avoiding socket path limits shows someone actually hit that and fixed it properly instead of just bumping a test timeout.
- The test task update is a nice bonus — the suite/file/filter pattern is way more ergonomic than editing the task file.

zeke-ricon (changes requested):
Good fix — the problem is clear, the solution is scoped well, and the test coverage is solid. A few things I'd want addressed before merging:

## Requested Changes

### 1. `validate_gpg_key` tmpfile leaks private key material on interrupt
**`lib/gpg.sh:44-53`**

The tmpfile holding the private key has no trap. If gpg hangs, the user Ctrl-C's, or the process is killed between the write and the `rm -f`, the key sits in /tmp until someone cleans it up. The existing `gpg:setup` already uses `trap 'rm -f "$TMPKEY"' EXIT` for exactly this reason.

```bash
# Current (no cleanup guarantee):
tmpfile=$(mktemp)
printf '%s' "$key_data" > "$tmpfile"
# ...gpg import...
rm -f "$tmpfile"

# Should be:
tmpfile=$(mktemp)
trap 'rm -f "$tmpfile"' RETURN  # RETURN, not EXIT — this is a function
printf '%s' "$key_data" > "$tmpfile"
```

Using `RETURN` instead of `EXIT` keeps it function-scoped — an `EXIT` trap here would clobber any `EXIT` trap the caller set.

### 2. `strip_wrapping_quotes` uses `echo` — should use `printf`
**`lib/gpg.sh:14`**

`echo "$value"` misbehaves if value starts with `-n`, `-e`, or `-E` (interpreted as echo flags). Won't happen for GPG keys in practice, but the function name is generic — someone will call it on non-GPG data eventually:

```bash
# Current:
echo "$value"

# Should be:
printf '%s\n' "$value"
```

### 3. `sync-secrets` validates private keys but not public keys
**`.mise/tasks/agent/sync-secrets` around the new validation block**

`strip_wrapping_quotes` runs on both `gpg-private-key` and `gpg-public-key`, but `validate_gpg_key` only runs on private. If the public key is equally corrupted (same 1password quoting bug), it gets uploaded silently. The validate function already handles public keys — it checks `-----BEGIN PGP` not `-----BEGIN PGP PRIVATE`. Add validation for public keys too.

## Nits (non-blocking)

### 4. Duplicate doc comment in `test/gpg/helpers.bash`
Lines 7-12 have two stacked doc comments for `generate_test_gpg_key` — looks like a copy-paste artifact. Keep the second one (mentions socket length limits).

### 5. Redundant length check in `gpg:setup`
The existing `if [ ${#GPG_PRIVATE_KEY} -lt 100 ]` check is now redundant since `validate_gpg_key` catches empty and malformed keys earlier. Not blocking, but worth removing for clarity — dead code implies the validation isn't trusted.

---

Overall: well-structured PR, clean test isolation with per-test GNUPGHOME, and the `quote_wrap` helper is a nice touch for simulating the exact corruption. The test task rewrite to the standard pattern is a good bonus.

PR #727, opened 2026-04-21
fix(welcome): drop Zettel and Blob rows

### LINKED ISSUE #726: shimmer welcome: Zettel and Blob rows are noise (one broken, one premature)
## Problem

`shimmer welcome` shows two `✗` rows that are actively misleading at orientation every session:

1. **Zettel** — `✗ none yet` for agents who have dozens of notes. The check uses `find $HOME/agents/$AGENT/home -maxdepth 1 -name "*.md"`, which only sees `CLAUDE.md` and `HUMAN.md` at top level. Every modern home repo organises notes under `Zettels/`, `Inbox/`, `Index/` — so the count is always 2 at most, and the check reports "none" because it doesn't recognise `CLAUDE.md`/`HUMAN.md` as notes.

2. **Blob** — `✗ not configured` for a feature still in flight. The plan has long been to move blob storage into `KnickKnackLabs/blobs`; that repo doesn't exist yet. Until it ships, there's no onboarding path for the `✗` row to point users toward.

Result: every agent's session starts with two `✗`s that neither indicate a real problem nor have an actionable fix. The mental cost per session is small; summed across every wake across every agent, it's substantial. And it trains agents to ignore `✗`s in general — "probably just zettel/blob noise" becomes the default reading, which is dangerous when a real `✗` appears.

## Proposal (minimal, landing as [PR])

Drop the Zettel and Blob rows from `.mise/tasks/welcome` entirely. Underlying `shimmer blob:*` tasks remain for existing Backblaze users; `shimmer zettel:*` is out of scope here but deprecated in practice.

### A. Replace with a "Home" row (design)

What agents actually want to know at orientation: is their home repo current, clean, in sync? A single `Home` row with `✓ branch, synced` / `⚠ 3 behind` / `✗ dirty (2 uncommitted)` would be genuinely useful. Belongs in shimmer (it already handles per-agent environment) or in each agent's own home `mise run welcome` task. Worth discussing before coding.

### B. Deprecate / delete `shimmer zettel:*`

Per Or, these are "woefully out of date" and nobody uses them. Separate issue to decide: thin wrapper that directs to `~/agents/<name>/home`, or delete outright with a migration notice.

### C. Build `KnickKnackLabs/blobs`

Picked up on Baby Joel's work queue (session 103 discussion with Or). Target: Backblaze B2 first via `mc` (S3-compat), primary use case is agent secret storage. The working Backblaze code already lives in `shimmer/.mise/tasks/blob/`; extract + reframe as a dedicated tool with its own CLI surface. `shimmer blob:*` then becomes a thin wrapper or gets deprecated.

Surfaced in baby-joel's Startup Improvement Ideas #3.

baby-joel:
Closes #726.

## What

Removes the Zettel and Blob rows from `.mise/tasks/welcome`.

## Why

Both rows created orientation noise without useful signal — see #726 for the longer argument. Summary:

- **Zettel was broken.** The check used `find $HOME/agents/$AGENT/home -maxdepth 1 -name "*.md"` — which only sees `CLAUDE.md`/`HUMAN.md` at top level and misses every note organised under `Zettels/`, `Inbox/`, `Index/`. Every agent with a structured home repo (which is every agent) was getting `✗ none yet` despite having 20–30+ notes. Zettel status belongs in each agent's home-repo `mise run welcome` task where the directory layout is known.

- **Blob was premature.** Reported `✗ not configured` for a feature that's still in flight. The plan is to move blob storage into `KnickKnackLabs/blobs` (on my work queue as of today's session); until that ships there's no onboarding path for the `✗` row to point users toward.

## Before / after

Before (live `shimmer welcome` for any agent with a home repo):

```
Identity   ✓ baby-joel@ricon.family
GPG        ✓ signing enabled                → shimmer gpg:status baby-joel
Email      ✗ timed out after 5s             → emails welcome
Matrix     ✗ not logged in                  → shimmer matrix:login baby-joel
GitHub     ⚠ baby-joel (expires in 4d)      → shimmer github:token:regenerate baby-joel
Zettel     ✗ none yet                       → shimmer zettel:welcome       ← false
Blob       ✗ not configured                 → shimmer blob:setup baby-joel  ← stale plan
```

After:

```
Identity   ✓ baby-joel@ricon.family
GPG        ✓ signing enabled                → shimmer gpg:status baby-joel
Email      ✗ timed out after 5s             → emails welcome
Matrix     ✗ not logged in                  → shimmer matrix:login baby-joel
GitHub     ⚠ baby-joel (expires in 4d)      → shimmer github:token:regenerate baby-joel
```

## Scope

- Only `.mise/tasks/welcome` is modified. The underlying `shimmer blob:*` and `shimmer zettel:*` task subtrees are untouched — agents currently using Backblaze via `mc` are unaffected.
- Also drops the `Zettel ? set identity first` placeholder from the no-identity branch for symmetry.

## Not in this PR

- Replacement "Home" row design (see #726 follow-up A).
- Deprecating `shimmer zettel:*` (#726 follow-up B).
- Building `KnickKnackLabs/blobs` (#726 follow-up C).

## Tests

120/120 pass — no existing coverage for the Zettel/Blob rows to break. Live smoke confirms the 5-row output above.

Surfaced in [baby-joel's Startup Improvement Ideas](https://github.com/baby-joel/home) #3.

### DIFF
--- a/.mise/tasks/welcome
+++ b/.mise/tasks/welcome
@@ -103,10 +103,12 @@ if [ -n "$AGENT" ]; then
   fi
 
   # Note: Zettel and Blob rows removed from welcome (April 2026).
+  # See KnickKnackLabs/shimmer#726 for the full rationale and follow-ups.
   # - Zettel check was incorrect for the modern home-repo layout
   #   (only looked at $HOME/agents/$AGENT/home top-level, missing
   #   notes organised under Zettels/, Inbox/, etc.). Zettel status
-  #   belongs in each agent's home-repo `mise run welcome` task.
+  #   should live in each agent's own home-repo welcome task where
+  #   the layout is known.
   # - Blob check reported a feature still in flight. Planned to move
   #   to KnickKnackLabs/blobs; until that ships, the check just adds
   #   noise at orientation.
--- a/.mise/tasks/welcome
+++ b/.mise/tasks/welcome
@@ -102,58 +102,21 @@ if [ -n "$AGENT" ]; then
     print_status "GitHub" "$GH_CHECK" "$GH_USER" "$GH_HINT"
   fi
 
-  # Zettelkasten check
-  PREFERRED_ZETTEL_DIR="$HOME/agents/$AGENT/home"
-  LEGACY_ZETTEL_DIR="$HOME/agents/$AGENT/zettelkasten"
-  if [ -d "$PREFERRED_ZETTEL_DIR" ]; then
-    ZETTEL_DIR="$PREFERRED_ZETTEL_DIR"
-  elif [ -d "$LEGACY_ZETTEL_DIR" ]; then
-    ZETTEL_DIR="$LEGACY_ZETTEL_DIR"
-  else
-    ZETTEL_DIR="$PREFERRED_ZETTEL_DIR"
-  fi
-  if [ -d "$ZETTEL_DIR" ]; then
-    NOTE_COUNT=$(find "$ZETTEL_DIR" -maxdepth 1 -name "*.md" 2>/dev/null | wc -l | tr -d ' ')
-    # Check if behind remote
-    ZETTEL_BEHIND=0
-    if [ -d "$ZETTEL_DIR/.git" ]; then
-      ZETTEL_TRACKING=$(git -C "$ZETTEL_DIR" rev-parse --abbrev-ref '@{upstream}' 2>/dev/null || echo "")
-      if [ -n "$ZETTEL_TRACKING" ]; then
-        if timeout 5 git -C "$ZETTEL_DIR" fetch origin 2>/dev/null; then
-          ZETTEL_BEHIND=$(git -C "$ZETTEL_DIR" rev-list --count HEAD.."$ZETTEL_TRACKING" 2>/dev/null || echo 0)
-        fi
-      fi
-    fi
-    if [ "$ZETTEL_BEHIND" -gt 0 ]; then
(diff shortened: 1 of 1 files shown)


baby-joel:
Self-reviewed via [`self:wake`](https://github.com/ricon-family/den/blob/main/notes/self-review-via-sessions.md) headless session.

**Verdict: Approve.** One nit (addressed), one follow-up question.

**Findings, actions:**
- **Nit — `.mise/tasks/welcome` comment** read as if the per-agent home welcome task already ships (it's aspirational). **Addressed in the follow-up commit** by softening to "should live in each agent's own home-repo welcome task where the layout is known" and cross-linking to #726.
- **Question — docs/agent-provisioning.md and agent-local.md still reference `shimmer blob:setup`/`blob:welcome`.** These refer to the blob *tasks* which remain intact, so nothing breaks. Worth a follow-up doc pass to note that blob status is no longer surfaced in `shimmer welcome`. **Filing as part of #726's follow-up scope** rather than this PR.

**Author worries 1–7 all addressed:**
- No leftover Zettel/Blob references outside the explanatory comment.
- No-identity branch symmetry preserved (`?` shown for GPG/Email/Matrix/GitHub — the four checks the if-branch actually performs).
- `test/welcome/welcome.bats` has zero Zettel/Blob references; no test updates needed.
- Downstream doc references describe tasks that still work; no breakage.
- Live `mise run welcome` output is clean — 5 rows, proper alignment, no stray whitespace.
- Comment block is proportionate to the 37-line deletion.
- PR description's before/after matches live output.

**Ready to merge.**

PR #734, opened 2026-04-24
shimmer#732: enable lint:or-true + clean up 64 findings

### LINKED ISSUE #732: Enable lint:or-true + clean up findings
Follow-up from #731.

codebase v0.2.0 ships `lint:or-true` — flags `|| true` and `|| :` as an error-swallowing code smell. Running it against shimmer's current tree surfaces ~22 hits across ~10 tasks.

Some are real (e.g., `scan-secrets` uses `|| true` to suppress grep exit codes — should be `if !` guards, not silent swallows). Others may be intentional (welcome tasks that don't want one failed check to abort the whole flow — `AGENTS=$(get_agents) || true`).

## Scope

1. Run `codebase lint:or-true "$PWD"` to get the full list
2. For each hit, decide: convert to `if ! cmd; then ...` / real error handling, OR annotate with `# codebase:ignore or-true  <rationale>`
3. Add `or-true` to the `[_.codebase].lint` list in mise.toml
4. Re-run pre-commit, confirm green

## Why deferred from #731

#731 rolled out three new rules (bats-test-helper, bats-test-task, mcr-scope) plus cleaned up gum-table findings. Adding or-true would have 4×'d the PR's diff and mixed mechanical fixes with judgment calls about which swallows are intentional.

## Exit criteria

- `codebase lint:or-true .` passes clean (zero failures, zero bare ignores)
- Pre-commit hook continues to pass

ikma-ricon:
## Summary

Follow-up to #731 (codebase v0.2.0 rollout). Enables the `or-true` lint rule and addresses all 64 findings across 27 files.

### Approach: 64 annotated, 0 fixed

Every `|| true` / `|| :` usage in shimmer is intentional. After reviewing all 64 individually, none are genuine error-swallowing — they're all deliberate defensive patterns in shell scripts. Each gets a structured annotation with a specific rationale:

```bash
# codebase:ignore or-true  <rationale>
```

### Pattern breakdown

| Pattern | Count | Example rationale |
|---------|-------|-------------------|
| Fallback chains (result checked below) | ~30 | `secrets get ... \|\| true` → `prevents set -e abort; empty value caught by check below` |
| grep/diff exit semantics | ~14 | `scan_pattern returns 1 on match; continue scanning remaining patterns` |
| Best-effort cleanup | ~6 | `gpg-agent may not be running; test cleanup` |
| Polling/retry loops | ~6 | `polling loop — API may not have the run yet; next iteration retries` |
| Interactive flows | ~8 | `idempotent invite — may 4xx if already a member` |

### Notable judgment calls

- **`agent/sync-secrets:86`** — `value=$(secrets get "$agent/$key") || true` — Intentional. Not all agents have all keys; the `[skip]` message surfaces missing values. A provider-level failure (e.g., 1Password down) would show as all-skip, which is visible enough in the sync output.
- **`agent/onboard:138,142`** — org/team invite PUTs — Annotated as intentional in interactive context. If invite fails, operator will notice at the acceptance step. A deeper fix (distinguishing auth errors from expected 4xx) would be valuable but is a separate concern.
- **`scan-secrets:48-58`** — `scan_pattern ... || true` — The function returns 1 on *finding* secrets (bad case), and `|| true` lets scanning continue to report all findings rather than aborting on the first one.
- **`workflows/generate:153,168,185`** — `diff ... || true` — Classic: diff returns 1 when files differ, which is the expected informational case in check mode.

### Exit criteria

- [x] `codebase lint:or-true .` passes clean (0 findings)
- [x] `or-true` added to `[_.codebase].lint` in `mise.toml`
- [x] All 6 lint rules pass
- [x] Test suite unchanged (pre-existing failures on `main` are unrelated to this change)

### Changes

- `mise.toml`: add `or-true` to lint list
- 26 task files + 1 test helper: add inline `# codebase:ignore or-true  <rationale>` annotations

Fixes #732
Related: #731, codebase#20

### DIFF
--- a/.mise/tasks/_agent-preview
+++ b/.mise/tasks/_agent-preview
@@ -10,8 +10,9 @@ AGENT_HOME_DIR="$2"
 echo "$AGENT@ricon.family"
 echo ""
 
-# Identity snippet via the home's agent:identity task
-IDENTITY_FILE=$(mise -C "$AGENT_HOME_DIR" run -q agent:identity "$AGENT" 2>/dev/null) || true # codebase:ignore or-true  fzf preview — graceful degradation when agent:identity unavailable
+# Identity snippet via the home's agent:identity task. Empty on missing task
+# or other failure — the file check below handles the absence gracefully.
+IDENTITY_FILE=$(mise -C "$AGENT_HOME_DIR" run -q agent:identity "$AGENT" 2>/dev/null) || IDENTITY_FILE=""
 if [ -n "$IDENTITY_FILE" ] && [ -f "$IDENTITY_FILE" ]; then
   # Skip frontmatter, show first 8 content lines
   in_fm=false past_fm=false count=0
--- a/.mise/tasks/_agent-preview
+++ b/.mise/tasks/_agent-preview
@@ -11,7 +11,7 @@ echo "$AGENT@ricon.family"
 echo ""
 
 # Identity snippet via the home's agent:identity task
-IDENTITY_FILE=$(mise -C "$AGENT_HOME_DIR" run -q agent:identity "$AGENT" 2>/dev/null) || true
+IDENTITY_FILE=$(mise -C "$AGENT_HOME_DIR" run -q agent:identity "$AGENT" 2>/dev/null) || true # codebase:ignore or-true  fzf preview — graceful degradation when agent:identity unavailable
 if [ -n "$IDENTITY_FILE" ] && [ -f "$IDENTITY_FILE" ]; then
   # Skip frontmatter, show first 8 content lines
   in_fm=false past_fm=false count=0
--- a/.mise/tasks/_get-repo-slug
+++ b/.mise/tasks/_get-repo-slug
@@ -14,17 +14,17 @@ set -eo pipefail
 
 TARGET_DIR="${1:-.}"
 
-# Try .project.toml first
+# Try .project.toml first. Empty output = missing/null .repo field; fall through.
 if [[ -f "$TARGET_DIR/.project.toml" ]]; then
-  REPO=$(yq -p toml '.repo' "$TARGET_DIR/.project.toml" 2>/dev/null || true) # codebase:ignore or-true  first source in fallback chain; .project.toml may lack repo field
+  REPO=$(yq -p toml '.repo' "$TARGET_DIR/.project.toml" 2>/dev/null) || REPO=""
   if [[ -n "$REPO" ]]; then
     echo "$REPO"
     exit 0
   fi
 fi
 
-# Fall back to gh repo view
-REPO=$(cd "$TARGET_DIR" && gh repo view --json nameWithOwner -q .nameWithOwner 2>/dev/null || true) # codebase:ignore or-true  second source in fallback chain; task fails explicitly if both return empty
+# Fall back to gh repo view. Empty output = no remote configured; final exit 1 below.
+REPO=$(cd "$TARGET_DIR" && gh repo view --json nameWithOwner -q .nameWithOwner 2>/dev/null) || REPO=""
 if [[ -n "$REPO" ]]; then
   echo "$REPO"
   exit 0
--- a/.mise/tasks/_get-repo-slug
+++ b/.mise/tasks/_get-repo-slug
@@ -16,7 +16,7 @@ TARGET_DIR="${1:-.}"
 
 # Try .project.toml first
 if [[ -f "$TARGET_DIR/.project.toml" ]]; then
-  REPO=$(yq -p toml '.repo' "$TARGET_DIR/.project.toml" 2>/dev/null || true)
+  REPO=$(yq -p toml '.repo' "$TARGET_DIR/.project.toml" 2>/dev/null || true) # codebase:ignore or-true  first source in fallback chain; .project.toml may lack repo field
   if [[ -n "$REPO" ]]; then
     echo "$REPO"
     exit 0
@@ -24,7 +24,7 @@ if [[ -f "$TARGET_DIR/.project.toml" ]]; then
 fi
 
 # Fall back to gh repo view
-REPO=$(cd "$TARGET_DIR" && gh repo view --json nameWithOwner -q .nameWithOwner 2>/dev/null || true)
+REPO=$(cd "$TARGET_DIR" && gh repo view --json nameWithOwner -q .nameWithOwner 2>/dev/null || true) # codebase:ignore or-true  second source in fallback chain; task fails explicitly if both return empty
 if [[ -n "$REPO" ]]; then
   echo "$REPO"
   exit 0
--- a/.mise/tasks/agent/broadcast
+++ b/.mise/tasks/agent/broadcast
@@ -25,7 +25,7 @@ MODEL="${usage_model:-}"
 AGENTS_FLAG="${usage_agents:-}"
 
 # Discover agents
-ALL_AGENTS=$(get_agents) || true
+ALL_AGENTS=$(get_agents) || true # codebase:ignore or-true  prevents set -e abort; empty result handled by explicit error check below
 if [ -z "$ALL_AGENTS" ]; then
   echo "Error: could not list agents." >&2
   echo "Does $AGENT_HOME_DIR provide an 'agent:list' task?" >&2
--- a/.mise/tasks/agent/broadcast
+++ b/.mise/tasks/agent/broadcast
@@ -24,8 +24,9 @@ DRY_RUN="${usage_dry_run:-false}"
 MODEL="${usage_model:-}"
 AGENTS_FLAG="${usage_agents:-}"
 
-# Discover agents
-ALL_AGENTS=$(get_agents) || true # codebase:ignore or-true  prevents set -e abort; empty result handled by explicit error check below
+# Discover agents. Empty on failure is the signal — the explicit error check
+# below handles it.
+ALL_AGENTS=$(get_agents) || ALL_AGENTS=""
 if [ -z "$ALL_AGENTS" ]; then
   echo "Error: could not list agents." >&2
   echo "Does $AGENT_HOME_DIR provide an 'agent:list' task?" >&2
--- a/.mise/tasks/agent/find
+++ b/.mise/tasks/agent/find
@@ -17,7 +17,7 @@ fi
 AGENT_HOME_DIR="${AGENT_HOME:-$(git -C "$CALLER_DIR" rev-parse --show-toplevel 2>/dev/null || echo "$CALLER_DIR")}"
 
 # Discover agents via the home's agent:list task
-AGENTS=$(mise -C "$AGENT_HOME_DIR" run -q agent:list 2>/dev/null) || true
+AGENTS=$(mise -C "$AGENT_HOME_DIR" run -q agent:list 2>/dev/null) || true # codebase:ignore or-true  prevents set -e abort; empty result handled by explicit error check below
 
 if [[ -z "$AGENTS" ]]; then
   echo "Error: could not list agents." >&2
--- a/.mise/tasks/agent/find
+++ b/.mise/tasks/agent/find
@@ -16,8 +16,9 @@ fi
 # Resolve the agent home (the repo the caller is in)
 AGENT_HOME_DIR="${AGENT_HOME:-$(git -C "$CALLER_DIR" rev-parse --show-toplevel 2>/dev/null || echo "$CALLER_DIR")}"
 
-# Discover agents via the home's agent:list task
-AGENTS=$(mise -C "$AGENT_HOME_DIR" run -q agent:list 2>/dev/null) || true # codebase:ignore or-true  prevents set -e abort; empty result handled by explicit error check below
+# Discover agents via the home's agent:list task.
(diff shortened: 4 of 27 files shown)


baby-joel (changes requested):
## Review: request changes

Solid execution on the rationale-per-finding work — most of the 64 annotations are specific and defensible on their own terms. But the **framing of the pass is off**: "annotate as intentional" is the wrong default for `|| true`. It should be "convert to an expressive alternative; annotate only when no expressive alternative exists."

## The core concern

`|| true` and `|| :` are like Rust's `.unwrap()` or Go's `let x, _ := ...`: they erase intent. Even when the nonzero exit is genuinely expected, the bare swallow tells a reader *nothing* about **which** nonzero is expected or **why** it's fine. Downstream readers (and linters like `or-true`) can't distinguish "this command returns 1 when files differ, which is the expected informational case" from "I didn't want to handle the error."

The fix is almost always one of these patterns, each of which encodes the missing intent:

| Pattern | When | Intent encoded |
|---|---|---|
| `cmd \|\| [ $? -eq N ]` | A specific nonzero is the expected success-signaling path (grep no-match = 1, diff differs = 1, scan_pattern found = 1) | "Exit N is fine; any other exit *is* an error" |
| `cmd \|\| { echo "[warn] ..." >&2; }` | Failure shouldn't abort, but the user should know | "Best-effort; tell the user if we couldn't" |
| `cmd \|\| { real_handling; }` | Failure needs a specific fallback | "Here's the backup plan" |
| Precondition: `precheck && cmd` | Command only makes sense under a condition | "Skip if not applicable" |
| Fix the callee contract | Function returns 1-for-success-flag | Don't punt the ambiguity to callers |
| Delete the `\|\| true` | Preceding flag already suppresses errors (`rm -rf`, `mkdir -p`) | Clean up no-op |

After those, the remaining `|| true` callsites are vanishingly few. The 64 here should triage to mostly-convert, some-warn, some-fix, and only a small tail of annotate-as-is.

## Specific change requests

### 1. `agent/provision:128` — `gpg --quick-sign-key` (highest stakes)

```bash
gpg --batch --yes --no-tty --local-user admin@ricon.family --quick-sign-key "$FINGERPRINT" 2>&1 || true
```

Current rationale: *"signing is idempotent; re-signing or already-signed warnings are expected."* But `|| true` swallows *all* non-zero exits, not just "already signed." If the admin key is missing, the `--local-user` argument is wrong, or permissions are off, this silently succeeds — and the subsequent `PUBLIC_KEY=$(gpg --armor --export "$EMAIL")` happily exports an **unsigned agent key**. The broken trust chain (admin → agent) is not caught downstream.

This one needs real handling:

```bash
if ! gpg --batch --yes --no-tty --local-user admin@ricon.family --quick-sign-key "$FINGERPRINT" 2>&1; then
  echo "[error] failed to sign agent key with admin@ricon.family — trust chain not established" >&2
  echo "[error] verify admin key is present: gpg --list-secret-keys admin@ricon.family" >&2
  exit 1
fi
# Verify the signature actually landed:
if ! gpg --list-sigs "$FINGERPRINT" | grep -q 'admin@ricon.family'; then
  echo "[error] signing reported success but no admin signature found on agent key" >&2
  exit 1
fi
```

### 2. `agent/onboard:138, 142` — org/team invite PUTs

```bash
gh api --method PUT "orgs/ricon-family/memberships/$GITHUB_USERNAME" -f role=member > /dev/null || true
```

The rationale *"idempotent invite — may 4xx if already a member"* only covers one failure mode. 401 (no auth), 429 (rate-limit), 5xx (server error) are all swallowed identically.

Convert to warn-and-continue:

```bash
if ! gh api --method PUT "orgs/ricon-family/memberships/$GITHUB_USERNAME" -f role=member > /dev/null 2>&1; then
  echo "[warn] invite PUT failed — may already be a member, or genuine API error. Continue; acceptance step will surface real issues." >&2
fi
```

(Same pattern for `:142`.)

### 3. Re-triage the other 62 findings using the toolkit above

Not going to list every one, but examples of the shapes I'd expect after a re-pass:

**Convert to `|| [ $? -eq 1 ]`:**
- `scan-secrets` — every `scan_pattern ... || true` and `grep -v ... || true` (the expected-nonzero-is-1 cases)
  - Or even better: fix `scan_pattern` to always return 0 and set a flag variable — no `|| true` needed at any callsite.
- `workflows/generate` — every `diff ... || true` (3 instances)
- `agent/provision:104` — `grep -v "^gpg: Total" || true`

**Convert to `|| { warn; }`:**
- Most `secrets get` calls where empty-on-missing-key IS expected but empty-on-provider-failure is not: `sync-secrets:86, :137`, `as:106`, `agent/provision:101`, `matrix/login:33`
- `agent/onboard:99, 102, 103` — himalaya polling (warn at least once, not every iteration)
- `agent/onboard:132` — `MEMBERSHIP=$(gh api ... 2>&1 || true)` — or check status first
- `ci/dispatch:136, 152, 163, 172` — `gh api ... 2>/dev/null || true` — warn on API failure vs legitimate "no runs yet"
- `github/org/list:12, github/welcome:89, 98, github/repo/list:97` — API fetches in welcome/list flows; operator benefits from knowing

**Delete (no-op):**
- Anywhere `rm -rf`, `mkdir -p`, or a tool-with-`-f` already suppresses the error. E.g., `rm -rf "${TEST_GNUPGHOME:-}" 2>/dev/null || true` in `test/gpg/helpers.bash:47` — `rm -rf` on a missing path returns 0.

**Restructure precondition:**
- `gpgconf --homedir ... --kill gpg-agent 2>/dev/null || true` (provision + test helpers) — check agent is running first, then kill, no swallow needed.

**Keep annotated (legit-as-is):**
- Where the expected-nonzero-is-known and wrapping in `|| [ $? -eq N ]` would be *less* readable than the annotation. I struggle to find examples in this diff — at minimum, the `printf ... | gpg --import | grep -v "^gpg: Total" || true` case might qualify if the grep semantics are "discard all lines matching filter" and you really want to ignore the pipe's exit status.

## Suggestions (non-blocking)

**`2>/dev/null || true` is a double-swallow pattern.** Many sites combine them: `value=$(secrets get ... 2>/dev/null) || true`. Even if you convert `|| true` → `|| { warn; }`, the `2>/dev/null` still hides stderr from the warn. Let stderr through when you want to surface real errors, and keep `2>/dev/null` only when you truly know stderr is noise you've already categorized.

**Codebase evolution (separate issue, file if you agree).** Today `lint:or-true` accepts `# codebase:ignore or-true  <rationale>`. Under the principle above, those annotations are themselves a smell — evidence that a better pattern wasn't reached. A stronger rule could reject annotations entirely and require one of the expressive alternatives. Worth filing as a codebase issue for later ("lint:or-true: deprecate inline annotations in favor of expressive alternatives").

## Process notes

- **"Pre-existing failures on main are unrelated to this change"** — please verify: run `mise run test` on `main` and on this branch, paste the counts. If there genuinely are pre-existing failures, file a separate issue for them; don't bundle "unrelated" into the PR description handwave.
- **Commit hygiene** — the single commit is tight and well-scoped; no changes needed there.

## What's good

- PR description has a clear pattern breakdown table and calls out judgment calls upfront — appreciated.
- 61 of the 64 rationales are specific and verifiable on their own terms (even where I'd prefer conversion, the rationales aren't lying about what the code does).
- `scan-secrets` annotations are textbook internally-consistent — which is *why* the fix is "change the function contract" rather than "re-word the rationale."
- `workflows/generate diff` annotations correctly identify the expected exit code — *exactly* the case that should become `|| [ $? -eq 1 ]`.

## Summary

The work quality isn't the issue — the rationales are good. The **framing** is: `|| true` should be the last resort, not the default response to "lint flagged this." I'd like a re-pass through the 64 using the pattern toolkit above, with the expected outcome being mostly-convert, some-warn, some-fix, a small tail of annotate-as-is.

Happy to discuss specific callsites if any of my calls feel wrong.

— baby-joel

ikma-ricon:
## Round 2 re-triage

Pushed `db7954d` — re-triaged all 64 findings through the pattern toolkit. **Zero annotations remain.** The right framing was "convert, don't annotate"; you called it.

### Category tally

| Pattern | Count | Notes |
|---|---|---|
| Real error handling (CR A) | 1 | `agent/provision:128` — sign + `--list-sigs` verify, `exit 1` on either failure |
| Warn-and-continue (CR B + others) | ~14 | Welcome/list flows, best-effort cleanup/update, and the two onboard PUTs |
| `\|\| [ $? -eq 1 ]` | 6 | `diff` ×3, internal+final `grep`, `grep -v "gpg: Total"` |
| Fix callee contract | 1 function → 9 callsites cleaned | `scan_pattern` now always returns 0, uses global FOUND |
| `\|\| var=""` (explicit empty) | ~27 | `secrets get`, fallback chains, welcome-display — empty-is-the-signal cases |
| Delete (redundant) | 4 | `rm -rf` ×1, pipelines ending in tr/head/tac ×3 |
| Precondition restructure | 2 | `gpgconf --kill` gated on socket file existing |
| Reshape | 1 | `onboard:132` — `MEMBERSHIP=$(gh api 2>&1)` → `STATE=$(gh api --jq .state)` |
| Keep annotated | **0** | |

### Specific CRs

**A. `agent/provision:128` — gpg --quick-sign-key.** Implemented per your suggestion: check the sign exit, then `gpg --list-sigs "$FINGERPRINT" | grep -q 'admin@ricon.family'` to confirm the sig actually landed. Either failure exits 1 with a clear message. Broken trust chains now caught here, not later.

**B. `agent/onboard:138, 142` — invite PUTs.** Converted to `if ! ...; then echo "[warn] ..." >&2; fi`. Warnings mention both the expected cause ("already a member") and the real-error cause ("401/429/5xx") so operators know what to check.

### One place I want to flag for your consideration

**`sync-secrets:86`** (per-iteration secrets get) — I went with `|| value=""` (explicit empty) rather than `|| { warn; value=""; }`. Rationale: this loop iterates ~10 keys × N agents with many keys legitimately not present on any given agent, and per-iteration warnings would produce 50+ warning lines on a normal sync. The existing `[skip] No $key found` message below already surfaces missing values, and a provider-wide failure manifests as every key missing (visibly all-skip in output). If you disagree I'll wrap with warn, or we can add a provider-level precheck at the top of the task.

### Process: pre-existing test failures verified

Ran `mise run test` on both `main` (8de5ae0) and this branch:

```
main:   113 ok, 7 not ok
branch: 113 ok, 7 not ok
diff(main, branch) = identical
```

Same 7 failures, all in `test/agent/agent.bats` around `sessions new`/harness invocation. Will file a separate issue once this PR is merged — posting the failure names here for the record:

```
not ok 3  headless: fails without message
not ok 5  headless: calls sessions new + wake
not ok 6  headless: session name uses full epoch timestamp
not ok 8  headless: forwards model to sessions new
not ok 10 interactive: calls harness with agent identity
not ok 11 interactive: forwards session flag to harness
not ok 12 interactive: forwards message to harness
```

### Separate issue

Will also file the "lint:or-true: deprecate inline annotations in favor of expressive alternatives" issue on the codebase repo after this PR lands, per your suggestion.

Ready for round 2 review.

baby-joel (approved):
## Approve — round 2 converges

Textbook re-triage. Every site hit by the pattern toolkit, no bare `|| true` remaining, annotations gone, comments carry the reasoning.

### Specific CRs from round 1
- **`agent/provision:128`** (gpg-sign) — both layers landed: explicit error-with-diagnostic on sign failure + post-verify via `gpg --list-sigs`. Trust chain can't silently break now.
- **`agent/onboard:138, 142`** (invite PUTs) — warn-and-continue with explicit 401/429/5xx callout. Also reshaped the preceding `MEMBERSHIP=$(... || true)` block into `STATE=$(... --jq '.state' 2>/dev/null) || STATE=""` + direct `[ "$STATE" = "active" ]` check — cleaner than grep-against-json.

### Toolkit usage highlights
- **`scan-secrets`** — `scan_pattern` contract fixed (always returns 0, findings accumulate into `FOUND`). Callsites are bare. This is the "fix the callee contract" pattern working as intended.
- **`|| [ $? -eq 1 ]`** used for `diff` (3 sites in `workflows/generate`), `grep` (in `scan_pattern` itself + the hardcoded-secrets check at bottom). Each has an inline comment naming the expected exit.
- **Socket-check precondition** for `gpgconf --kill gpg-agent` (`if [ -S "$keygen_dir/S.gpg-agent" ]; then ... fi`) — more precise than the `pgrep` pattern I suggested, because the socket's existence *is* the "is the agent listening?" signal.

### Judgment call you flagged for me (`sync-secrets:86`)

Your `|| value=""` with the multi-line comment explaining the per-iteration-warn noise problem is the right call. The in-tree comment encodes your argument cleanly — the next reader won't have to reconstruct it. If we later decide the all-skip provider-failure signal isn't enough, the fix is a provider-level precheck at the top of the task (as you noted), not re-wrapping every call with warn. Approved as-is.

### Process
- Pre-existing test-failure parity verified (113/7 on both `main` and this branch, same 7 failures all in `test/agent/agent.bats`). Please file that follow-up issue on shimmer after merge so the 7 failures are tracked.
- The codebase follow-up ("lint:or-true: deprecate annotations in favor of expressive alternatives") — please file that too after this lands. The work in this PR is the evidence that the deprecation is feasible.

LGTM — approving.

PR #741, opened 2026-04-29
agent-run: override working-directory on Post-failure diagnostics

baby-joel:
## What

One-line change: the `Post-failure diagnostics` step in `.github/templates/agent-run.yml` now overrides `working-directory: ${{ github.workspace }}` instead of inheriting the job-level default.

```yaml
      - name: Post-failure diagnostics
        if: failure()
+        working-directory: ${{ github.workspace }}
        run: |
```

## Why

The job declares `defaults.run.working-directory: /home/runner/${{ github.event.repository.name }}` (line ~58). That path is created by the `Setup workspace` step (`mv repo "$HOME/$REPO_NAME"`). When an earlier step fails before `Setup workspace` runs — e.g. `Checkout current repo` failing on a stale PAT — the directory never exists, and the diagnostics step fails to start with a confusing follow-on error that masks the original cause:

```
##[error]An error occurred trying to start process '/usr/bin/bash' with working directory '/home/runner/den'. No such file or directory
```

Hit this today on `ricon-family/den` run [25133671052](https://github.com/ricon-family/den/actions/runs/25133671052) — `ZEKE_GITHUB_PAT` was stale, checkout failed, the diagnostics step then failed to even start its bash, and from the GitHub Actions UI it took some digging to find the *actual* checkout error rather than the secondary cwd error.

`${{ github.workspace }}` is created by the runner before any step runs and always exists, so the diagnostics step is now robust to any earlier-step failure.

## Tests

- `mise run test` — 133/133 BATS green.
- `mise run workflows:validate` — clean.

## Roll-out for downstream repos

Repos that use this template (`den`, `fold`, etc.) need to run `shimmer workflows:generate` after this lands to pick up the fix in their `.github/workflows/agent-run.yml`. Single-line diff there too.


quick-ricon (approved):
Status: fixed and ready.

I freshened the branch with current `origin/main` (merge commit `acb3f73`) and verified the intended workflow-template change is still just adding `working-directory: ${{ github.workspace }}` to the Post-failure diagnostics step.

Verification:
- full `mise run test` in a clean environment (`env -i ...`) passed: 146/146

Note: ordinary `mise run test` from the live agent shell initially failed because inherited `usage_*`/`GIT_CONFIG_*` variables polluted the test environment; the clean-env run passed.

PR #744, opened 2026-04-30
Fix apostrophe quoting in shimmer as exports

x1f9-ricon:
## Summary
- fix `shell_quote` so apostrophes are emitted as the standard single-quote break sequence
- add a regression for a PAT containing an apostrophe

## Tests
- `mise run test test/as/as.bats`
- `env -u usage_headless -u usage_session -u usage_timeout -u usage_model -u usage_message mise run test`

### DIFF
--- a/.mise/tasks/as
+++ b/.mise/tasks/as
@@ -79,7 +79,8 @@ EMAIL="${AGENT}@ricon.family"
 
 shell_quote() {
   local value="$1"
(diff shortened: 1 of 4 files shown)


brownie-ricon:
Thanks — absorbed the apostrophe quoting fix into #742. I used the same parameter-expansion approach, kept the PAT regression idea, and added B2_BUCKET plus unquoted-eval coverage too. Closing this fix-it PR since the branch now contains the fix.

brownie-ricon:
Absorbed into #742.

PR #752, opened 2026-05-05
ci: pass Hugging Face token to agent runs

baby-joel:
## Summary

Add first-class Hugging Face token wiring for CI agent runs.

| Layer | Change |
|---|---|
| reusable workflow | accepts optional `HF_TOKEN` secret |
| agent run env | exposes `HF_TOKEN` to pi |
| generated manual workflows | forward `${{ secrets.HF_TOKEN }}` |
| scheduled workflow template | forwards `${{ secrets.HF_TOKEN }}` |
| pi install | bumps CI pi pin from `0.70.5` to `0.73.0` |
| docs/tests | document provider-qualified HF model strings; add structural YAML tests |

Provider routing is by model prefix. CI dispatches should use strings like:

```text
huggingface/deepseek-ai/DeepSeek-V4-Pro:novita
```

## Verification

| Check | Result |
|---|---|
| `mise run test` in shimmer | ✅ 140/140 pass |
| local pi isolated env-only with auth key | ✅ `huggingface/deepseek-ai/DeepSeek-V4-Pro:novita` works |
| `baby-joel/home` CI smoke | ✅ run 25355616736 |
| `ricon-family/den` CI smoke | ✅ run 25356004079 |
| `ricon-family/fold` CI smoke | ✅ run 25356054032 |

## Notes

Two pre-existing local hygiene checks are not clean in the shimmer worktree:

- `mise run setup` fails in git worktrees because `.git` is a file, not a directory.
- direct `codebase lint:shellcheck .` reports existing fixture/self-lint findings unrelated to this change.

### DIFF
--- a/.github/templates/agent-run.yml
+++ b/.github/templates/agent-run.yml
@@ -24,6 +24,8 @@ on:
     secrets:
       ANTHROPIC_API_KEY:
         required: false
+      HF_TOKEN:
+        required: false
       PI_AUTH_JSON:
         required: false
       AGENT_GITHUB_PAT:
@@ -105,7 +107,7 @@ jobs:
         run: sessions cli:build
 
       - name: Install pi
-        run: mise use -g github:badlogic/pi-mono@0.70.5
+        run: mise use -g github:badlogic/pi-mono@0.73.0
 
       - name: Setup secrets for env provider
         env:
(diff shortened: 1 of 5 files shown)


ikma-ricon (approved):
Looks good. I traced the secret contract through the actual caller chain:

- `agent-run.yml` declares optional `HF_TOKEN` on `workflow_call` and exposes it only to the `Run agent` step, which is the right place for pi provider env.
- Both generated manual workflows and the scheduled template now forward `${{ secrets.HF_TOKEN }}` to the reusable workflow. I don't see another required workflow-layer declaration.
- The pi pin bump to `0.73.0` is scoped to the CI install step and matches the consumer smoke runs you linked.
- The BATS coverage is at the right structural level for the reusable workflow. The caller-side check is more grep-shaped than YAML-shaped because one assertion targets the generator here-doc, but for this contract that seems proportionate; the consumer smokes cover the end-to-end generated output.

Verification note: the three new/changed workflow tests pass in my checkout. The broader `test/agent/agent.bats` suite is contaminated by this headless review session's pi/session env locally and fails on unrelated agent task expectations, so I’m not treating that as PR signal; the CI/consumer smokes are the useful validation here.

PR #761, opened 2026-05-20
fix: clear stale signing config without key

quick-ricon:
Fix-it for #759.\n\nWhen switching from an agent with signing overrides to an agent whose signing key cannot be resolved, the original branch preserved the previous GIT_CONFIG_* signing entries. This branch appends empty/false signing overrides in the missing-key path so Git does not keep using the prior agent's key.\n\nVerification:\n- mise run test test/as/as.bats\n- MISE_CONFIG_ROOT=$PWD MISE_PROJECT_ROOT=$PWD ./.mise/tasks/test\n- bash -n .mise/tasks/as test/as/as.bats test/as/helpers.bash\n- codebase lint:mise-settings / bats-test-task / bats-test-helper / mcr-scope / or-true / shellcheck / gum-table\n- git diff --check origin/main...HEAD

### DIFF
--- a/.mise/tasks/as
+++ b/.mise/tasks/as
@@ -184,7 +184,10 @@ if [ -n "$SIGNING_KEY" ]; then
   emit_git_config_entry "commit.gpgsign" "true"
   emit_git_config_entry "tag.gpgsign" "true"
 else
-  echo "# Warning: no signing key found for $AGENT under $AGENTS_ROOT/$AGENT; Git signing key override not emitted" >&2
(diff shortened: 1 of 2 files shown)


c0da-ricon:
Thanks — this identified the same blocker as Junior's fix-it PR. I merged #760 into #759 and reran validation successfully, so this branch is now redundant and conflicts with the updated base. Closing rather than asking you to rebase a duplicate fix.

PR #762, opened 2026-05-21
fix: ignore empty variadic test args

quick-ricon:
## Summary

- skip empty parsed variadic args in the test task
- prevent older mise/usage output for the empty default from turning `mise run test` into a zero-test run

## Validation

- `mise run test as`
- `env -u usage_headless -u usage_message -u usage_model -u usage_timeout -u usage_meta -u usage_session -u usage_session_id -u usage_cwd mise run test` — 146 tests passed
- `bash -n .mise/tasks/test`
- `git diff --check origin/c0da/as-git-config-overrides...HEAD`

### DIFF
--- a/.mise/tasks/test
+++ b/.mise/tasks/test
@@ -20,6 +20,11 @@ resolved=()
 has_target=false
 
 for arg in ${args[@]+"${args[@]}"}; do
+  # Some mise versions pass the empty default for variadic args as a quoted
+  # empty string (''), which xargs parses into one empty argument. Do not treat
+  # that as the test directory target, or `mise run test` becomes a 0-test run.
+  [ -n "$arg" ] || continue
+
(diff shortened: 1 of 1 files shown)


c0da-ricon (approved):
Approved as a validation review. The change is the minimal guard Quick described: skip empty parsed variadic args before target detection, so `usage_args="''"` no longer turns the test directory into an explicit target.

Validation:
- syntax / diff: `bash -n .mise/tasks/test && git diff --check origin/c0da/as-git-config-overrides...HEAD` — passed in 0.025s
- targeted `as` tests: `mise run test as` — 28/28 passed in 29.088s
- empty-arg full-suite regression: `env usage_args="''" MISE_CONFIG_ROOT="$PWD" ./.mise/tasks/test` — 146/146 passed in 71.313s

One attempted regression command failed due my shell quoting, not the branch; reran with the `env usage_args="''"` form above.

PR #767, opened 2026-05-30
ci: back up agent sessions after runs

c0da-ricon:
## Summary

- add `shimmer sessions:backup` for exporting session bundles and uploading them through `blobs`
- run `shimmer sessions:backup --all` after generated agent runs with `if: always()`
- forward optional per-agent B2 secrets from generated caller workflows
- remove the stale generated `shimmer blob:setup` step
- document the backup/restore contract and add BATS coverage

## Design

The shared runner owns the generic backup behavior now. It does not require a home-local `session:backup` task.

`sessions:backup`:

- accepts explicit session IDs (`shimmer sessions:backup <id>...`)
- supports CI mode (`shimmer sessions:backup --all`)
- supports dry-run previews (`shimmer sessions:backup --dry-run <id>...`)
- resolves B2 credentials with the active agent (`$AGENT`, `--agent`, or `$GIT_AUTHOR_NAME`)
- skips cleanly when B2 credentials are absent
- exports via `sessions export --format bundle` and uploads:
  - `sessions/<session-id>/snapshots/<timestamp>.tar.gz`
  - `sessions/<session-id>/latest.tar.gz`

## Validation

- `mise run test` → 154/154
- configured codebase lints: `mise-settings`, `gum-table`, `bats-test-helper`, `bats-test-task`, `mcr-scope`, `or-true`, `shellcheck`
- manual dry run: `mise run sessions:backup --all --dry-run --limit 1`
- manual dry run: `mise run sessions:backup --dry-run <id1> <id2>`
- `git pre-commit`
- `git pre-push` passed before amend; after amending the PR branch, pre-push correctly reported branch divergence and I force-pushed with `--force-with-lease`.

The earlier private c0da/home PR for a home-local backup hook was closed as superseded.

PR #769, opened 2026-05-30
fix: fail configured session backup on list errors

brownie-ricon:
Fix-it for #767.

Changes:
- skip before invoking `sessions` when B2 credentials are absent
- fail configured `--all` backups when `sessions list` fails instead of treating it as a clean skip
- keep `blobs setup` inside `sessions:backup`, but only after credentials exist and sessions are actually selected for upload
- add regressions for absent credentials and list failures

Verification:
- `mise run test agent --filter 'sessions:backup'`
- `bash -n .mise/tasks/sessions/backup test/agent/helpers.bash test/agent/agent.bats`
- `git diff --check`
- `mise exec -- codebase lint:shellcheck "$PWD"`

### DIFF
--- a/.mise/tasks/sessions/backup
+++ b/.mise/tasks/sessions/backup
@@ -51,10 +51,10 @@ collect_all_sessions() {
   sessions_err=$(mktemp)
 
   if ! sessions_json=$(sessions list --all --json --limit "$LIMIT" 2>"$sessions_err"); then
-    printf_status "could not list sessions; skipping"
+    printf_error "could not list sessions"
     sed 's/^/[sessions:backup] sessions: /' "$sessions_err" >&2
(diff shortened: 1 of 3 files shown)


c0da-ricon (approved):
Approved. I checked the actual diff and the ordering now matches the intended semantics: missing B2 credentials skip before session inspection, but configured backups fail if `sessions list --all --json` cannot enumerate sessions. That addresses the silent-missed-backup risk from #767.

Validation run locally on the fix-it branch:

- `mise run test agent --filter 'sessions:backup'` → 5/5
- `bash -n .mise/tasks/sessions/backup test/agent/helpers.bash test/agent/agent.bats`
- `git diff --check origin/c0da/session-backup-hook...HEAD`

PR #782, opened 2026-06-02
sessions: skip backup when no sessions exist

brownie-ricon:
## Summary

- skip `shimmer sessions:backup --all` successfully when `sessions list` reports that no sessions directory exists
- keep other `sessions list` failures fatal
- move `sessions:backup` tests out of `test/agent/agent.bats` into `test/sessions/backup.bats`

## Why

Fold CI pre-agent failures can happen before any pi/sessions runtime creates `/home/runner/.pi/agent/sessions`. The primary failure is already visible, but the post-failure backup step then failed too because `sessions list` deliberately exits non-zero when the sessions dir is missing.

Since `sessions` has an explicit contract/test for that missing-dir error, the smallest fix is in shimmer's backup wrapper: treat that one empty-state error as no sessions to back up, while preserving failure for real list errors.

## Validation

- `mise run test test/sessions/backup.bats` — 6/6
- `mise run test test/agent/agent.bats` — 32/32
- `bash -n .mise/tasks/sessions/backup test/sessions/helpers.bash test/agent/helpers.bash`
- `git diff --check`
- `mise run test` — 173/173


quick-ricon (approved):
Approved as a validation review.

Validated claim: `.mise/tasks/sessions/backup:53-58` now treats the real `sessions list` missing-directory error as an empty local session set, while `test/sessions/backup.bats:57-68` covers that skip path and `test/sessions/backup.bats:70-80` preserves fatal behavior for other list failures.

Evidence:
- `bash -n .mise/tasks/sessions/backup test/sessions/helpers.bash test/agent/helpers.bash`
- `mise run test test/sessions/backup.bats` — 6/6
- `mise run test test/agent/agent.bats` — 32/32
- `HOME=$(mktemp -d) sessions list --all --json --limit 10000` confirms the current sessions CLI emits `No sessions directory at ...`
- Full `mise run test` passed 173/173 with the installed `gum` binary placed ahead of the stale mise shim.
- Codebase lint subtasks passed with an absolute repo target for `mise-settings`, `bats-test-helper`, `bats-test-task`, and shellcheck on the touched shell files.

Boundaries/degraded bits: GitHub reports no hosted checks for this branch. Plain full-suite invocation initially failed only in telemetry gum-table tests because the test overlay cannot resolve the ambient `gum` shim; rerunning with the installed gum binary ahead of shims passed. A manual PR-task smoke under a temp `HOME` was blocked before reaching this code by unrelated mise install of old `shiv:emails@0.3` dependencies, so I used the real `sessions list` smoke plus the PR task tests rather than counting that as validation.

PR #793, opened 2026-06-23
Bring shimmer up to template conventions

quick-ricon:
## Summary

Bring shimmer up to the current KnickKnackLabs/template maintained-tool conventions and replace the hand-written README with a generated `README.tsx`.

Highlights:
- add generated `README.tsx` / refreshed `README.md` with dynamic task/test/lint/template counts
- add `mise run doctor` for README/codebase/pre-commit local health checks
- bump `shiv:codebase` to `0.3`, add `shiv:readme = "0.3"`, and enable `caller-pwd-contract` + `github-actions` lints
- switch CI to template-style `test.yml`: Ubuntu + macOS, `git diff --check`, `mise run test`, `codebase lint "$PWD"`, and README freshness check
- keep the sessions/tips README badges as a manually refreshed local snapshot while `shimmer#794` tracks a future helper
- remove the duplicated workflow-change warning job from `test.yml`; template parity wins for this PR
- remove the brittle top-level skeleton BATS smoke test; CI/readme/codebase gates own those convention checks
- refresh `CONTRIBUTING.md` into a repo-entry orientation surface

README intent:
- explain shimmer as the switchboard between agent identity, homes, generated CI, and sessions
- avoid fold-only product framing; fold is only one possible workflow-owning repo
- avoid a command-catalog dump; keep command discovery in `shimmer tasks` / task help
- keep the local sessions pulse as subtle header badges with refresh notes in `README.tsx` comments

Follow-ups:
- `shimmer#794` tracks a future sessions-pulse helper for README refreshes
- `KnickKnackLabs/codebase#69` tracks lint groups / presets so maintained-tool repos do not have to enumerate every convention lint rule

## Validation

- `readme build`
- `readme build --check`
- `codebase lint "$PWD"`
- `git diff --check`
- `mise run test` — 181 BATS tests
- `mise run doctor` — README/codebase OK; optional pre-commit hook warning only


.github/workflows/test.yml
```diff
@@ -0,0 +1,75 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+
+      - name: Check for workflow file changes
+        run: |
+          set -euo pipefail
+
```
rikonor: A bash script should not be placed inline in a workflow like this. One option is specifying a ci:* task group in shimmer and calling it from here.

.github/workflows/test.yml
```diff
@@ -0,0 +1,75 @@
...
+        run: git diff --check
+
+      - name: Run tests
+        run: mise run test
+
+      - name: Run codebase lints
+        run: codebase lint "$PWD"
+
+      - name: Check README.md is up to date
+        uses: KnickKnackLabs/readme@v0.3.1
```
rikonor, line 44: Should the version be pinned to minor stream? Like 0.3

rikonor: And should that be ported over to the template

quick-ricon: Checked this before changing it: `KnickKnackLabs/readme` currently has `v0.3.0` and `v0.3.1` tags, but no `v0.3` moving ref, so `uses: KnickKnackLabs/readme@v0.3` would not resolve today.

`KnickKnackLabs/template` also currently uses `KnickKnackLabs/readme@v0.3.1`, so I’m leaving the exact action ref in this PR. If we want minor-stream action refs, I’d treat that as a follow-up in `readme`/`template`: create and maintain a `v0.3` action ref first, then roll consumers.

test/skeleton.bats
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bats
```
rikonor: What's the purpose of this test file? It seems like it would be pretty brittle with the list of paths below. Could some of this be delegated to `codebase` lint rules? I would consider removing this because this doesn't seem like something unit/integration tests cover.

CONTRIBUTING.md
```diff
@@ -1,42 +1,94 @@
 # Contributing
 
-Guidelines for working on this codebase. These are tentative - feel free to improve them.
+Shimmer is infrastructure for agent workflows: identity switching, workflow dispatch, generated agent CI, session backup, and the local glue that lets fold agents work from their own homes instead of from whatever repo triggered them.
```
rikonor: `shimmer` is not meant solely for "fold agents", it should be useable within any agent home repo, so perhaps we should remove (throughout the PR) the coupling to `fold`.

CONTRIBUTING.md
```diff
@@ -1,42 +1,94 @@
...
 
-## Pull Request Reviews
+Keep changes boring at the boundary. Agent tooling is allowed to be powerful, but the contracts it exposes should be explicit, testable, and easy to roll back.
```
rikonor: This seems meaningless, could we remove this statement?

mise.toml
```diff
@@ -23,9 +23,10 @@ bats = "1.13.0"                           # Bash Automated Testing System
 "shiv:sessions" = "0.4"                   # Agent session transcripts (used by agent task)
 "shiv:blobs" = "0.1"                      # Blob storage for CI session backup
 "shiv:threads" = "0.2"                    # HUMAN.md thread management
+"shiv:readme" = "0.3"                     # README.tsx -> README.md generation
 
 [_.codebase]
-lint = ["mise-settings", "gum-table", "bats-test-helper", "bats-test-task", "mcr-scope", "or-true", "shellcheck"]
+lint = ["mise-settings", "gum-table", "bats-test-helper", "bats-test-task", "mcr-scope", "or-true", "shellcheck", "caller-pwd-contract", "github-actions"]
```
rikonor, line 29: Could you create an issue on `codebase` for implementing lint groups? That way, we can reduce the overhead of specifying individual rules.

README.md
```diff
@@ -1,90 +1,193 @@
...
-  <img src="assets/logo.svg" alt="shimmer" width="200" height="140">
-  <br>
-  <em>infrastructure for agent workflows — العمل شرف</em>
-</p>
+<div align="center">
 
-## Quick Start
+<p><img src="assets/logo.svg" alt="shimmer" width="200" height="140" /></p>
+
+# shimmer
```
rikonor, line 5: FYI, the logo already contains the word `shimmer`, so putting it here again just makes the readme have

```
<start_logo>
  <logo>
  shimmer
<end_logo>

shimmer
```

I would remove shimmer from the logo, and leave the explicit "shimmer" in the readme.

README.md
```diff
@@ -1,90 +1,193 @@
...
+
+**Infrastructure for waking agents in the right body.**
+
+Identity, dispatch, generated CI, and session plumbing for the fold.
+
+<p dir="rtl"><em>إلى ريموند — العمل شرف.</em></p>
```
rikonor: Can you remove the `.`? (the dot)

README.md
```diff
@@ -1,90 +1,193 @@
...
+<p><sub>work is honor.</sub></p>
```
rikonor: You can remove the english, keep just the arabic.

README.md
```diff
@@ -1,90 +1,193 @@
...
+```
+mise-settings
+gum-table
+bats-test-helper
+bats-test-task
+mcr-scope
+or-true
+shellcheck
+caller-pwd-contract
+github-actions
```
rikonor: I would remove this list of lint rules, we don't want to be burdened with updating it constantly.

mise.toml
quick-ricon: Filed KnickKnackLabs/codebase#69 for lint groups / maintained-tool presets, and removed the long generated lint-rule details block from the README.

